PR #13, opened 2019-12-16
create app

quinn:


pkg/apply/apply.go
```diff
@@ -10,15 +11,37 @@ import (
...
 	cmd := exec.Command("kubectl", "apply", "-f", "-")
 	stdin, err := cmd.StdinPipe()
-	defer stdin.Close()
 
 	if err != nil {
 		return
 	}
 
 	stdin.Write(bytes)
+	stdin.Close()
```
jkminneti: Just curious, why no defer? Does the system exit cause issues or something?

pkg/events/filter.go
```diff
@@ -18,9 +19,9 @@ func filter(e *util.RegistryEvent) (event *pendingRelease) {
 	branch := slicedTag[2]
 
 	if name == "tuber" && branch == "master" {
-		return &pendingRelease { name: name, branch: branch }
-	} else {
-		fmt.Println("Ignoring", name, branch)
+		return &pendingRelease{name: name, branch: branch}
```
jkminneti: Oh right lol

pkg/apply/apply.go
quinn: the stream needs to be closed before CombinedOutput can be called.

pkg/events/filter.go
quinn: the linter made me do it

pkg/events/filter.go
```diff
@@ -2,25 +2,34 @@ package events
...
 }
 
-func filter(e *util.RegistryEvent) (event *pendingRelease) {
-	filterRegex := regexp.MustCompile(`us\.gcr\.io\/(.*):(.*)`)
-	slicedTag := filterRegex.FindStringSubmatch(e.Tag)
-	name := slicedTag[1]
-	branch := slicedTag[2]
+func filter(e *util.RegistryEvent) (event pendingRelease, err error) {
+	apps, err := k8s.TuberApps()
+	var matchApp k8s.TuberApp
```
jkminneti: if this was a pointer, this function _could_ look less like javascript
```go
for _, app := range apps {
	if app.ImageTag == e.Tag {
		matchApp = &app
		break
	}
}

if matchApp == nil {
	fmt.Println("Ignoring", e.Tag)
	return
}
```
just an option

PR #14, opened 2019-12-16
add ack, fix regex

jkminneti:


pkg/events/stream.go
```diff
@@ -1,17 +1,26 @@
...
-func Stream(ch chan *util.RegistryEvent) {
+func Stream(ch chan *util.RegistryEvent, token string) {
 	for event := range ch {
-		event := filter(event)
-		if event != nil {
-			token := os.Getenv("GCLOUD_TOKEN")
-			release.New(event.name, event.branch, token)
+		pendingRelease := filter(event)
+		if pendingRelease == nil {
+			return
```
quinn: i think we should Ack here

PR #17, opened 2019-12-17
cache the response

quinn:


pkg/k8s/configmap.go
```diff
@@ -80,3 +87,19 @@ func TuberApps() (apps []TuberApp, err error) {
...
 	return
 }
+
+// TuberApps returns a list of tuber apps
+func TuberApps() (apps []TuberApp, err error) {
+	if cache == nil || cache.timestamp < time.Now().Unix()-60*5 {
+		apps, err = getTuberApps()
+
+		cache = &appsCache{
+			apps:      apps,
```
jkminneti: so in the error case, it gets set to a non-nil pointer with an empty array tho

pkg/k8s/configmap.go
```diff
@@ -80,3 +87,19 @@ func TuberApps() (apps []TuberApp, err error) {
 
 	return
 }
+
+// TuberApps returns a list of tuber apps
+func TuberApps() (apps []TuberApp, err error) {
+	if cache == nil || cache.timestamp < time.Now().Unix()-60*5 {
```
jkminneti: maybe store the expiration time? also oof I hope there's a better way

PR #20, opened 2019-12-18
fix auth

jkminneti:


pkg/k8s/configmap.go
```diff
@@ -85,11 +85,13 @@ func getTuberApps() (apps []TuberApp, err error) {
...
 
-	for name, imageTag := range config.Data {
+	for _, imageTag := range config.Data {
 		split := strings.SplitN(imageTag, ":", 2)
+		repoSplit := strings.SplitN(split[0], "/", 2)
+		repoName := repoSplit[1]
 
 		apps = append(apps, TuberApp{
-			Name:     name,
+			Name:     repoName,
```
quinn: err, this should be the name. the name should be e.g. `tuber` so we can make `tuber-configmap` `tuber-secret-xyz` etc etc. its the prefix for everything in the cluster. Also, it has to be unique to the app, so we can have e.g. `tuber-demo-configmap`, and also `tuber-eng-configmap` (because the repoName will be the same, but the name will be different)

pkg/layers/registry.go
```diff
@@ -22,7 +22,7 @@ type registry struct {
 // NewGoogleRegistry creates registry struct
 func newGoogleRegistry(googleToken string) *registry {
 	return &registry{
-		baseURL:  "https://us.gcr.io",
```
quinn: this should probably be passed as a param, and we should extract it from the pubsub event, from `TuberApp.ImageTag`

pkg/k8s/configmap.go
jkminneti: well the object needs this new `repoName` somewhere - I can leave the name alone, but what'd be best? Add a new attribute for it, or replace `Repo`? Just not sure because we're not using `Repo` anywhere and idk your plans for it

pkg/listener/listen.go
```diff
@@ -64,7 +64,7 @@ func (l *listener) startListener(ctx context.Context) error {
 	// Set a fixed deadline and small outstanding message count for testing
 	// TODO: Remove or make configurable
 	subscription.ReceiveSettings.MaxExtension = cfg.AckDeadline
-	subscription.ReceiveSettings.MaxOutstandingMessages = 5
+	//subscription.ReceiveSettings.MaxOutstandingMessages = 5
```
jkminneti: @dbennett-freshly should I leave this in?

dbennett-freshly: I made in configurable later so I'll deal with the conflict after we merge this.

PR #22, opened 2019-12-19
Allow for configuration of debug mode and listener timeouts

dbennett-freshly:
This time there is no slash in the branch name.

This PR adds configuration for debug via a command-line switch and max-timeout / max-outstanding via environment variables (MAX_TIMEOUT and MAX_OUTSTANDING -- or something).

I guess I should have used the prefix thing so it'd be TUBER_MAX_OUTSTANDING but I didn't.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -25,7 +25,7 @@ func init() {
 	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
 
 	rootCmd.PersistentFlags().BoolP("debug", "d", false, "debug")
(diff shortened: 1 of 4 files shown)


cmd/root.go
```diff
@@ -1,21 +1,44 @@
...
 	rootCmd = &cobra.Command{
 		Use:   "tuber",
 		Short: "",
 	}
-
-	app string
 )
 
+func init() {
+	viper.AutomaticEnv()
```
dbennett-freshly: This ties the environment to Viper.

PR #24, opened 2019-12-20
Add fetch command consolidated k8s

dbennett-freshly:


pkg/containers/repository.go
```diff
@@ -155,7 +165,7 @@ func convertResponse(response *http.Response) (yamls []util.Yaml, err error) {
 		}
 
 		if !strings.HasPrefix(header.Name, ".tuber") {
-			err = &notTuberLayerError{"contains stuff other than .tuber"}
+			err = &notTuberLayerError{"contains stuff other than .pulp"}
```
quinn: WTF is "pulp"

dbennett-freshly: Okay, so there might have been a few too many replaces.. I fixed that.

PR #25, opened 2019-12-20
Fix the stream goroutine to avoid races / var overwrites

dbennett-freshly:


pkg/events/stream.go
```diff
@@ -21,11 +22,14 @@ func (s *streamer) Stream(chIn <-chan *util.RegistryEvent, chOut chan<- *util.Re
...
 	defer close(chErr)
 
+	var wait *sync.WaitGroup
+
 	for event := range chIn {
 		pendingRelease, err := filter(event)
 
 		if err != nil {
 			chErr <- err
+			chOut <- event
```
dbennett-freshly: It appears as though we never plan to not ack messages, right? If so, we should ack on errors like the above.

pkg/events/stream.go
```diff
@@ -34,22 +38,29 @@ func (s *streamer) Stream(chIn <-chan *util.RegistryEvent, chOut chan<- *util.Re
...
+				zap.String("releaseBranch", pendingRelease.Tag))
 
-			_, err = release.New(pendingRelease, s.token)
+			releaseLog.Info("release: starting")
+
+			_, err := release.New(pendingRelease, s.token)
 
 			if err != nil {
-				releaseLog.Warn("Release: error", zap.Error(err))
+				releaseLog.Warn("release: error", zap.Error(err))
```
dbennett-freshly: Log messages should be lowercase. I added these, I'm attempting to repay my debts.

pkg/events/stream.go
```diff
@@ -34,22 +38,29 @@ func (s *streamer) Stream(chIn <-chan *util.RegistryEvent, chOut chan<- *util.Re
 			continue
 		}
 
-		var releaseLog = s.logger.With(
-			zap.String("releaseName", pendingRelease.Name),
-			zap.String("releaseBranch", pendingRelease.Tag))
+		go func(event *util.RegistryEvent) {
```
quinn: why not use the closure var?

pkg/events/stream.go
quinn: i don't tihnk so, if there is an error it is not something we expected, and i don't think implies we want to "pass" on the event

pkg/events/stream.go
dbennett-freshly: Because it will change value as the loop iterates we need to bind a goroutine-specific var to he object we actually want to operate on.

pkg/events/stream.go
dbennett-freshly: Except we don't do anything with the errors?  Like, they just sit there and get re-errored repeatedly?

What should we do on an invalid message? It'll never become valid, right?

jkminneti:
Could be from the master merge, but I'm getting this:
```
go run main.go start
{"level":"info","ts":1577202972.4836512,"caller":"listener/listen.go:129","msg":"acknowledge loop: starting"}
{"level":"info","ts":1577202972.484385,"caller":"listener/listen.go:97","msg":"Listener: starting"}
{"level":"info","ts":1577202972.4844549,"caller":"cmd/start.go:43","msg":"error listener: started"}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x107a526]

goroutine 32 [running]:
sync.(*WaitGroup).state(...)
	/usr/local/go/src/sync/waitgroup.go:33
sync.(*WaitGroup).Add(0x0, 0x1)
	/usr/local/go/src/sync/waitgroup.go:54 +0x26
tuber/pkg/events.(*streamer).Stream.func1(0xc0003aa048, 0xc000222540, 0xc000387ea0, 0xc0000a2660, 0xc0000a21e0, 0xc0003374c0)
	/Users/jordanminneti/work/tuber/pkg/events/stream.go:42 +0x4f
created by tuber/pkg/events.(*streamer).Stream
	/Users/jordanminneti/work/tuber/pkg/events/stream.go:41 +0x1b0
exit status 2
error: Recipe `start` failed on line 5 with exit code 1
```

dbennett-freshly:
@jkminneti I pushed the fix for the wg nil pointer

jkminneti:
> @jkminneti I pushed the fix for the wg nil pointer

Yup, works, reviewing changes now

pkg/events/stream.go
```diff
@@ -21,35 +22,43 @@ func (s *streamer) Stream(chIn <-chan *util.RegistryEvent, chOut chan<- *util.Re
...
+			wait.Add(1)
+			defer wait.Done()
+
+			var err error
+			defer func() {
+				if err != nil {
+					chErr <- err
+				} else {
+					chOut <- event
+				}}()
```
jkminneti:
```suggestion
				}
			}()
```

PR #32, opened 2020-01-03
nack, ack, listen better

jkminneti:


.tuber/deployment.yaml
```diff
@@ -16,7 +16,7 @@ spec:
         app: tuber
     spec:
       containers:
-      - image: gcr.io/freshly-docker/tuber@sha256:238373034a9ee5f0eeddc89b6f2d40a43cbfcd13b84e290d4ce49a8f137c9955
+      - image: gcr.io/freshly-docker/tuber@sha256:bdbea73cb20e7888929705691949897a814122589257a45b05d3832a6b2ecad8
```
quinn: shouldn't this be master now, or the hack we talked about

jkminneti: yup, I was just doing a test deploy - switched it back to master

PR #33, opened 2020-01-03
patch config cmd

jkminneti:


cmd/config.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/spf13/cobra"
+)
+
+var configCmd = &cobra.Command{
+	Use: "config [patch || remove]",
+}
+
+var configPatchCmd = &cobra.Command{
+	Use:  "patch [appName] [key] [value]",
```
Hbbb: How do you feel about `set` and `unset`? Patch and remove don't feel like opposites to me.

PR #37, opened 2020-01-05
interpolate

jkminneti:
_dont_ run deploy to test this one (without a breakpoint at least) - no image override is present, as no interpolation is happening on the latest master `.tuber/deployment.yaml`.


pkg/events/updateImage.go
```diff
@@ -0,0 +1,92 @@
...
+	kind, err := dyno.GetString(convertedYaml, "kind")
+	if err != nil {
+		return
+	}
+	deployment = kind == "Deployment"
+	return
+}
+
+func updateDeployment(convertedYaml map[string]interface{}, digest string, filename string) (updatedYaml util.Yaml, err error) {
+	err = dyno.Set(convertedYaml, digest, "spec", "template", "spec", "containers", 0, "image")
```
jkminneti: actually - can I match on app name here? if the deployment has multiple containers I obviously can't use index 0

jkminneti:
branch name is deprecated i promise

pkg/events/interpolate.go
```diff
@@ -0,0 +1,33 @@
+package events
+
+import (
+	"sync"
+	"tuber/pkg/dataTemplate"
+)
+
+func interpolate(yamls []dataTemplate.Yaml, data map[string]string) (interpolated []dataTemplate.Yaml, err error) {
```
jkminneti: tomorrow: test `map[string]string` for `GetConfig` with an int value in the configmap - i still don't think it'll work. If it doesn't, and if we ever want to support a configmap per app for custom env-specific interpolatables, these should all become `map[string]interface{}` :(

pkg/events/publish.go
```diff
@@ -0,0 +1,24 @@
...
+	yamls, err := containers.GetTuberLayer(app.GetRepositoryLocation(), token)
+
+	if err != nil {
+		return
+	}
+
+	// TODO: make this smarter and separate
+	data := map[string]string{"tuberImage": event.Digest}
+
+	output, err = k8s.ApplyYamls(yamls, data, event.ContainerName())
```
quinn:
```suggestion
	output, err = k8s.ApplyYamls(yamls, data, app.Name)
```

pkg/k8s/configmap.go
```diff
@@ -36,7 +36,7 @@ func (c *Config) Save() (err error) {
 		return
 	}
 
-	write(jsondata)
+	Apply(jsondata, "tuber")
```
quinn: I don't think this should be hard-coded like this. when we run `tubectl env set appname VAR value` that will be in namespace `appname` not `tuber`

PR #38, opened 2020-01-06
Fix create command to have error messages

quinn:


cmd/list.go
```diff
@@ -19,20 +22,23 @@ to quickly create a Cobra application.`,
 	RunE: list,
 }
 
-func list(cmd *cobra.Command, args []string) error {
+func list(cmd *cobra.Command, args []string) (err error) {
```
jkminneti: this doesn't need to return anything anymore, if we're just fataling out

PR #40, opened 2020-01-07
fixin that create

quinn:


pkg/k8s/namespace.go
```diff
@@ -18,15 +22,66 @@ func CreateNamespace(namespace string) (err error) {
...
+		rules:
+		- apiGroups: ["", "extensions", "apps"]
+		  resources: ["*"]
+		  verbs: ["*"]
+		- apiGroups: ["batch"]
+		  resources:
+		  - jobs
+		  - cronjobs
+		  verbs: ["*"]
+		---
```
jkminneti: ah heck. see #37 - thought I had the only place where we'd want multiple yamls interpolated and converted to a `---` separated byte array

PR #41, opened 2020-01-07
POPS-232 Add Sentry

Hbbb:


cmd/start.go
```diff
@@ -82,6 +89,28 @@ func start(cmd *cobra.Command, args []string) {
...
+	if sentryEnabled {
+		err := sentry.Init(
+			sentry.ClientOptions{
+				Dsn:              viper.GetString("sentry-dsn"),
+				AttachStacktrace: true,
+			},
+		)
+
+		if err != nil {
+			panic(err)
```
Hbbb: Many error checks return error, should I actually panic here? Should the other error returns be converted to panics?

jkminneti: since we're adding sentry with the long-term goal of mostly picking up panics, I think panicking in `start` is the right way to go. 

None of this initialization stuff should ever error (in theory) so it should blow tf up. 

Long term Dan's right - sentry should be involved with _very_ few listener/events errors - basically only errors that the current code and config would introduce to all of an app's deploys (and any and all panics)

cmd/start.go
```diff
@@ -82,6 +89,28 @@ func start(cmd *cobra.Command, args []string) {
 	streamer := events.NewStreamer(token, logger)
 	go streamer.Stream(unprocessedEvents, processedEvents, failedEvents)
 
+	// Report any errors to Sentry
+	sentryEnabled := viper.GetBool("sentry-enabled")
+
+	if sentryEnabled {
```
jkminneti: I know this is exactly how my PR had it, but I think this should all be drop-in-replaceable. Like a separate package just for sentry, that probably takes the error channel and handles everything. Even to the point of  `errors.InitSentry(reportableErrors)`

cmd/start.go
```diff
@@ -67,7 +74,7 @@ func start(cmd *cobra.Command, args []string) {
 		return
 	}
 
-	unprocessedEvents, processedEvents, failedEvents, err := l.Listen(ctx, creds)
+	unprocessedEvents, processedEvents, failedEvents, reportableErrors, err := l.Listen(ctx, creds)
```
jkminneti: for some reason I feel like we should be creating the errors channel out here and passing it into the listener constructor. 

up to you though

pkg/listener/listen.go
```diff
@@ -21,7 +21,11 @@ type listener struct {
 	unprocessed chan *util.RegistryEvent
 	processed   chan *util.RegistryEvent
 	failures    chan util.FailedRelease
-	wait        *sync.WaitGroup
+
+	// TODO: What should this channel receive?
```
jkminneti: `error`, or a custom struct containing an error and metadata ---- but i think just `error` like you have it is ideal

cmd/start.go
jkminneti: actually, sentry initialization should happen before any functions that can return errors - yeah I think making this channel and initializing should happen riiiiight at the start here.

pkg/listener/listen.go
```diff
@@ -64,11 +70,11 @@ func NewListener(logger *zap.Logger, options ...Option) *listener {
 }
 
 // Listen for incoming pubsub requests
-func (l *listener) Listen(ctx context.Context, credentials []byte) (<-chan *util.RegistryEvent, chan<- *util.RegistryEvent, chan<- util.FailedRelease, error) {
+func (l *listener) Listen(ctx context.Context, credentials []byte) (<-chan *util.RegistryEvent, chan<- *util.RegistryEvent, chan<- util.FailedRelease, <-chan error, error) {
```
jkminneti:
```suggestion
func (l *listener) Listen(ctx context.Context, credentials []byte) (<-chan *util.RegistryEvent, chan<- *util.RegistryEvent, chan<- util.FailedRelease, chan<- error, error) {
```

cmd/start.go
```diff
@@ -82,6 +89,28 @@ func start(cmd *cobra.Command, args []string) {
...
+		if err != nil {
+			panic(err)
+		}
+
+		defer sentry.Recover()
+
+		for err := range reportableErrors {
+			alertSentry(err)
+		}
+	}
```
jkminneti: come to think of it, do we need to provide an internal receiver (fallback) for `reportableErrors` for when sentry isn't enabled?

like one that just logs?

Hbbb: Indeed we do. Good catch!

PR #42, opened 2020-01-08
kill util

jkminneti:
had to be done


pkg/listener/events.go
```diff
@@ -16,7 +14,7 @@ type ackable interface {
 }
 
 // FailedRelease can be created while streaming, consumed by the listener
-type FailedRelease interface {
-	Err() error
-	Event() *RegistryEvent
+type FailedRelease struct {
+	Err   error
+	Event *RegistryEvent
```
quinn: Is this correct? Did I fuck up the merge?

jkminneti: i mean, the build failed so maybe lol

PR #45, opened 2020-01-10
clusterwide interpolatables

jkminneti:


jkminneti:
no longer needed! though the refactor to have a single place for interpolation in core should probably still happen at some point

cmd/clusterData.go
```diff
@@ -0,0 +1,14 @@
+package cmd
+
+import (
+	"tuber/pkg/core"
+
+	"github.com/spf13/viper"
+)
+
+func clusterData() (data *core.ClusterData) {
```
Hbbb: Not sure that I like this in the `cmd` package, but I like it more than reading from the env in any other package. 🤔

jkminneti: only other thing I was thinking was a `cmd/helpers.go` for this kind of stuff 🤷‍♂ can do that later i guess if we need it

PR #48, opened 2020-01-22
Init Command

### LINKED ISSUE #49: Tuber Init Command
# Tuberize an app
* Create `.tuber` directory in project root
* Create `deployment.yaml`
* Create `service.yaml`

Hbbb:
* Creates `.tuber/` directory
* Creates a `deployment.yaml` and interpolates app name

@quinn wants us to use this package to embed static files in the binary, rather than hardcoding them in the source code. Thoughts? 

## TODO
* Don't generate the template with tabs (use pkger package)
* Actually test this

resolves #49


pkg/core/init.go
```diff
@@ -0,0 +1,146 @@
...
+	apiVersion: networking.istio.io/v1alpha3
+	kind: VirtualService
+	metadata:
+		name: {{.appName}}-ingress
+		namespace: {{.appName}}
+	spec:
+		hosts:
+			- "*"
+		gateways:
+		- istio-system/tls-gateway
```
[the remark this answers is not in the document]
Hbbb: But, aren't we doing the bare configuration repo? Or is that not happening anymore?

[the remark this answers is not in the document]
jkminneti: _we_ are, but unless the docs just say "make a bare config repo here are some yamls", it only works for us

cmd/init.go
```diff
@@ -0,0 +1,27 @@
...
+	"tuber/pkg/core"
+
+	"github.com/spf13/cobra"
+)
+
+var initCmd = &cobra.Command{
+	Use:   "init [appName] [routePrefix]",
+	Short: "initialize a .tuber directory and relevant yamls",
+	Long: `App name is the name of your app, which will be interpolated into the configuration files and used as the
+		namespace, as well as other things. Port number is the port that your app will run on. The standard is port 80.`,
```
jkminneti:
```suggestion
		namespace, as well as other things.`,
```

cmd/init.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"tuber/pkg/core"
+
+	"github.com/spf13/cobra"
+)
+
+var initCmd = &cobra.Command{
+	Use:   "init [appName] [routePrefix]",
+	Short: "initialize a .tuber directory and relevant yamls",
```
jkminneti: shouldn't all the stuff currently happening in `create` move to this?

pkg/core/init.go
```diff
@@ -0,0 +1,146 @@
...
+		- match:
+			- uri:
+				prefix: {{.routePrefix}}
+			route:
+			- destination:
+					host: {{.appName}}
+`)
+
+func Init(appName string, routePrefix string) (err error) {
+	if err = createTuberDirectory(); err != nil {
```
jkminneti: I feel like we're getting into a bit of trouble with this kind of thing - the error will probably be incredibly vague. I'm thinking we could `Errorf` a file-specific message, and concat in the actual error's message

pkg/core/init.go
```diff
@@ -0,0 +1,146 @@
...
+import (
+	"bytes"
+	"io/ioutil"
+	"os"
+	"text/template"
+
+	"github.com/MakeNowJust/heredoc"
+)
+
+const tuberConfigPath = ".tuber"
```
jkminneti: I like this idea, but the code that looks for that dir in the image should use the same source for that string (it's currently just right in the code over there)

cmd/init.go
```diff
@@ -0,0 +1,27 @@
...
+		namespace, as well as other things. Port number is the port that your app will run on. The standard is port 80.`,
+	Args: cobra.ExactArgs(2),
+	RunE: initialize,
+}
+
+func initialize(cmd *cobra.Command, args []string) (err error) {
+	appName := args[0]
+	routePrefix := args[1]
+
+	return core.Init(appName, routePrefix)
```
jkminneti: I'd say just throw the args right in, no reason to allocate locals. Very pedantic though lol

pkg/core/init.go
```diff
@@ -0,0 +1,146 @@
...
+
+var virtualService = heredoc.Doc(`
+	apiVersion: networking.istio.io/v1alpha3
+	kind: VirtualService
+	metadata:
+		name: {{.appName}}-ingress
+		namespace: {{.appName}}
+	spec:
+		hosts:
+			- "*"
```
jkminneti: Oh right ugh. This really shouldn't be there - I'll play around with my yamls tomorrow to avoid the star

jkminneti: ok yeah, can confirm that it works fine with our dns there instead. So that should probably be input

pkg/core/init.go
```diff
@@ -0,0 +1,146 @@
...
+const tuberConfigPath = ".tuber"
+
+var deployment = heredoc.Doc(`
+	apiVersion: apps/v1
+	kind: Deployment
+	metadata:
+		labels:
+			app: {{.appName}}
+		name: {{.appName}}
+		namespace: {{.appName}}
```
jkminneti: oh and I know we talked verbally about removing namespaces from these, but I think we should keep em there, just for explicitness

pkg/core/init.go
jkminneti: Oh right - it's env specific. So we do need "clusterwide interpolatables" still.

jkminneti: (https://github.com/Freshly/tuber/pull/45)

pkg/core/init.go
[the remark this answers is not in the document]
jkminneti: this would also be resolved by (https://github.com/Freshly/tuber/pull/45)

pkg/core/init.go
jkminneti: `{{ .clusterDefaultHost }}`

pkg/core/init.go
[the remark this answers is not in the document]
jkminneti: `{{ .clusterDefaultGateway }}` for now i think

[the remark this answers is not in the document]
Hbbb: 👍

cmd/init.go
Hbbb: No, this is like the `npm init` or `git init` command equivalent. It just creates the necessary files to get your app tuberized. It shouldn't actually deploy the app.

jkminneti: yeah fair

pkg/core/init.go
```diff
@@ -0,0 +1,146 @@
...
+		http:
+		- match:
+			- uri:
+				prefix: {{.routePrefix}}
+			route:
+			- destination:
+					host: {{.appName}}
+`)
+
+func Init(appName string, routePrefix string) (err error) {
```
jkminneti: Oh actually this name is scary - `func init` is reserved just like `main`

pkg/core/init.go
Hbbb: 👍

pkged.go
```diff
@@ -0,0 +1,8 @@
...
...269f3cd021cb52d2b3c1bbc7388bc342a364adc6fdf1ca941a0badca6ad4c05673cf9e79afa7e8b959a8e99a429150d43eaf34ec1a92618b1fb153659f4c5398d2ed3aab558deda739b95a48c39618ff0bc498a76bc4f60acc2037ff676f7274543b0fd188d936ace30d6c5d3a4fe191cfe633d2c14509f9f427f088b24979553699c88bd09c446dd3cbf81185d2d8d2ed20a7543b8f4d229e8118e248ba60b8e9789b12323d6b9ca97ea33e9ff02f24ed105ff0fdc90fd8d63a2a06f479845ae5ecef543d3b65ca98b5b56ae9bcd655dfdbb0e6b405bbee42b061f4ab644437c7f6c86dbadbbfd89ed3765f8aeec1d14fd80e470c3ca0e131a64c0fa023d6fea731442bcfd85c17f4cce63129bbcbda935f9f12818bc70f63faa7447a2822ee802627274687c1f30b6805b282a77796f4e9f1fdbbe72be8f489c6e456dcf9bd6f0d1cdc01cfefa42bb86a9f17eb88e16e9b7a7523bc7bde5c015f19a0b81073e3e3af46efa15c2a6eeea6e62ace3a3a9eae00fb0b9c25547b8fb324761416e7d59f7cbb61cc4917ee3c16a1c3ab8f45d0e406e42a2e5d84fc7197fd8e9ebb380c13b46e585d1046c3fad4dcd5bf7ccd2651289f4099733cc88dd12cbfd027caaf0245490f35e509ff9f76177dc91e1f1edf67dfbe7dbbcb8aa6c55fbfdd65153adca269b6689a2d9a668ba6d9a269b6689a2d9a668ba6d9a269b6689a2d9a668ba6d9a269b6689a2d9a668ba6d9a269b6689a2d9a668ba6d9a269b6689a2d9a668ba6d9a269b6689a2d9a668ba6d9a269b6689a2d9a668ba6d9a269b6689abf6c34cdff010000ffff010000ffffd649b1339b460000`)))
```
jkminneti: this doesn't have to be built, right?

pkged.go
```diff
@@ -0,0 +1,8 @@
+package main
+
+import (
+	"github.com/markbates/pkger"
+	"github.com/markbates/pkger/pkging/mem"
```
jkminneti: looks like the updated version is github.com/markbates/pkger

yamls/service.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{.appName}}
+  namespace: {{.appName}}
+spec:
+  ports:
+  - port: 3000
+    name: grpc
```
jkminneti: this should be configurable in some way

jkminneti:
can confirm the current state works, though

pkged.go
jkminneti: oops, meant to put this in `init.go`, looks like you're using both packages somehow?

cmd/init.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/spf13/cobra"
+)
+
+var initCmd = &cobra.Command{
+	Use:   "init [appName] [routePrefix]",
+	Short: "initialize a .tuber directory and relevant yamls",
+	Long: `App name is the name of your app, which will be interpolated into the configuration files and used as the
+		namespace, as well as other things.`,
+	Args: cobra.ExactArgs(2),
+	RunE: initialize,
```
jkminneti: `SilenceUsage: true,` with `RunE` pls 🙏 (new thing)

pkg/core/init.go
```diff
@@ -0,0 +1,97 @@
...
+		return
+	}
+
+	var buff bytes.Buffer
+
+	if err = tpl.Execute(&buff, templateData); err != nil {
+		return
+	}
+
+	path := "./tuber/" + fileName
```
jkminneti: This is wrong now, right? `.tuber`

Hbbb:
This is currently broken when you compile the binary with `go install` and run this from another directory. I think I just misunderstood how to use pkger.

PR #56, opened 2020-01-27
Make Pub/Sub Subscription Name Configurable

Hbbb:
Unsure of a few things here
* Should the default be enforced at the viper/env var level, or inside of the `NewListener` initializer? (I left both for now)
* Should this setting even use the whole `func WithSetting() Option` pattern, or does it deserve it's own argument space in the initializer?

I'm open to creating a "default" subscription inside `freshly-docker` and using that as the default value for this, but I'm partial to enforcing an env var to configure this. The risk is that if I deploy a new tuber to a different cluster, and the default is a subscription that another tuber uses, they will ack each other's messages.

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -73,6 +73,8 @@ func start(cmd *cobra.Command, args []string) {
 		options = append(options, listener.WithMaxTimeout(viper.GetDuration("max-timeout")))
 	}
 
+	}
+
 	var l = listener.NewListener(logger, options...)
 
 	creds, err := credentials()
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -2,10 +2,11 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"os"
 	"os/signal"
 	"syscall"
-	"tuber/pkg/errors"
+	sentry "tuber/pkg/errors"
 	"tuber/pkg/events"
 	"tuber/pkg/gcloud"
 	"tuber/pkg/listener"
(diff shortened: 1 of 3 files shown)


cmd/start.go
```diff
@@ -73,6 +73,10 @@ func start(cmd *cobra.Command, args []string) {
 		options = append(options, listener.WithMaxTimeout(viper.GetDuration("max-timeout")))
 	}
 
+	if viper.IsSet("pubsub-subscription-name") {
+		options = append(options, listener.WithSubscriptionName(viper.GetString("pubsub-subscription-name")))
```
quinn: can we document env vars in the readme? Like, this one would be:

```
TUBER_PUBSUB_SUBSCRIPTION_NAME
```

i think

Hbbb: Is `TUBER_` automatically prepended? Or are you suggesting I change it to be that?

quinn: ya its automatically prepended

cmd/start.go
```diff
@@ -2,10 +2,11 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"os"
 	"os/signal"
 	"syscall"
-	"tuber/pkg/errors"
+	sentry "tuber/pkg/errors"
```
jkminneti: not sure I agree with this, `errors.Stream` isn't necessarily sentry (as it's an integration)

jkminneti: Oh right - yeah it'd have to be `errorStream` I guess. alright up to you

Hbbb: In #58 I fixed the poor naming, so it's no longer necessary to alias this.

PR #59, opened 2020-02-02
Prerelease

jkminneti:


pkg/core/interpolate.go
```diff
@@ -14,3 +20,133 @@ func tuberData(app *TuberApp, digest string) (data map[string]string) {
...
+			continue
+		}
+	}
+	return []byte{}, fmt.Errorf("unhandled prerelease phase watch exit")
+}
+
+func checkPhase(name string, kind string, app *TuberApp) (out []byte, err error) {
+	cmd := exec.Command("kubectl", "get", kind, name, "-n", app.Name, "-o", `go-template="{{.status.phase}}"`)
+
+	out, err = cmd.CombinedOutput()
```
quinn: I think we need more unified / standardized handling of kubectl shell. And maybe, we should check the return code of the command there, and return an error at that point.

pkg/core/interpolate.go
```diff
@@ -14,3 +20,133 @@ func tuberData(app *TuberApp, digest string) (data map[string]string) {
...
+			return []byte{}, err
+		}
+
+		switch stringStatus := strings.Trim(string(status), `"`); stringStatus {
+		case "Succeeded":
+			return []byte{}, nil
+		case "Failed":
+			message, err := investigateFailure(name, kind, app)
+			if err != nil {
+				return message, err
```
quinn: i think we should combine err and message into a new error, and:
```
return nil, combinedErr
```

pkg/core/interpolate.go
```diff
@@ -14,3 +20,133 @@ func tuberData(app *TuberApp, digest string) (data map[string]string) {
...
+			}
+			return message, fmt.Errorf(string(message))
```
quinn: same here, return nil.

pkg/core/interpolate.go
```diff
@@ -14,3 +20,133 @@ func tuberData(app *TuberApp, digest string) (data map[string]string) {
...
+			}
+			return
+		}
+
+		return deletePrereleaser(prereleaser.Metadata.Name, kind, app)
+	}
+	return []byte{}, fmt.Errorf("unhandled prerelease run exit")
+}
+
+func waitForPhase(name string, kind string, app *TuberApp) ([]byte, error) {
```
quinn: it looks like `out` is only used in the case of an error, so i think this could just return an error

pkg/core/interpolate.go
```diff
@@ -14,3 +20,133 @@ func tuberData(app *TuberApp, digest string) (data map[string]string) {
...
+			return
+		}
+
+		out, err = waitForPhase(prereleaser.Metadata.Name, kind, app)
+		if err != nil {
+			deleteOut, deleteErr := deletePrereleaser(prereleaser.Metadata.Name, kind, app)
+			if deleteErr != nil {
+				doubleFailOut := []byte(string(out) + "\n also failed delete:" + string(deleteOut))
+				doubleFailErr := fmt.Errorf(err.Error() + "\n also failed delete:" + deleteErr.Error())
+				return doubleFailOut, doubleFailErr
```
quinn: same thing for this code. We are in an erroneous state, so we should only return the error. no need for `doubleFailOut`.

jkminneti: this answers most of these comments - harrison has a custom error constructor inside k8s that this would use once fully implemented

pkg/core/interpolate.go
```diff
@@ -14,3 +20,133 @@ func tuberData(app *TuberApp, digest string) (data map[string]string) {
...
+	Kind     string
+	Metadata Metadata
+}
+
+type Metadata struct {
+	Name string
+}
+
+func RunPrerelease(tubersTemp []string, app *TuberApp, digest string) (out []byte, err error) {
+	for _, tuber := range tubersTemp {
```
quinn: we should probably wrap everything in the for loop in a goroutine and use a `sync.WaitGroup`

jkminneti: PoC 🙏 true tho

jkminneti: though hmm... I still would say prerelease steps are an ordered thing - meaning not assumed to be a parallel thing

pkg/k8s/kubectl.go
```diff
@@ -24,58 +35,27 @@ func Apply(bytes []byte, namespace string) (out []byte, err error) {
...
-func Remove(name string, namespace string, data string) (out []byte, err error) {
-	cmd := exec.Command("kubectl", "patch", name, "-n", namespace, "--type=json", "-p", data)
-
-	out, err = cmd.CombinedOutput()
-
-	return
+// Create creates a resource with a given name and namespace
+func Create(namespace string, resourceAndArgs ...string) ([]byte, error) {
+	create := []string{"create", "-n", namespace}
+	return kubectl(append(create, resourceAndArgs...)...)
```
jkminneti: oops, this hack didn't work. fixing

pkg/k8s/errors.go
```diff
@@ -2,19 +2,38 @@ package k8s
...
+}
+
+// K8sError generic kubectl error, which also provides the error interface for AlreadyExistsError and NotFoundError
+type K8sError struct {
+	message string
+	err     error
+}
+
+func (e K8sError) Error() string {
+	return fmt.Sprintf("k8s: %s, %s", e.err.Error(), e.message)
```
jkminneti: `2020/02/04 17:44:04 k8s: exit status 1, Error from server (NotFound): namespaces "adsdf" not found`

pkg/core/interpolate.go
```diff
@@ -25,3 +30,82 @@ func ApplyTemplate(namespace string, templatestring string, params map[string]st
...
+// Metadata exported for the yaml unmarshaller
+type Metadata struct {
+	Name string
+}
+
+// RunPrerelease takes an array of pods, that are designed to be single use command runners
+// that have access to the new code being released.
+func RunPrerelease(tubers []string, app *TuberApp, digest string, clusterData *ClusterData) (out []byte, err error) {
+	for _, tuber := range tubers {
+		prereleaser := prerelease{}
```
quinn: here's an even worse option:

```
prerelease := prereleaseType{}
```


...... is this why go devs use single letter var names

pkg/core/interpolate.go
```diff
@@ -25,3 +30,82 @@ func ApplyTemplate(namespace string, templatestring string, params map[string]st
...
+		}
+
+		out, err = ReleaseTubers([]string{tuber}, app, digest, clusterData)
+		if err != nil {
+			return
+		}
+
+		out, err = waitForPhase(prereleaser.Metadata.Name, "pod", app)
+		if err != nil {
+			deleteOut, deleteErr := k8s.Delete("pod", prereleaser.Metadata.Name, app.Name)
```
quinn:
```suggestion
			_, deleteErr := k8s.Delete("pod", prereleaser.Metadata.Name, app.Name)
```

pkg/core/interpolate.go
```diff
@@ -25,3 +30,82 @@ func ApplyTemplate(namespace string, templatestring string, params map[string]st
...
+			if deleteErr != nil {
+				return deleteOut, fmt.Errorf(err.Error() + "\n also failed delete:" + deleteErr.Error())
```
quinn:
```suggestion
				return nil, fmt.Errorf(err.Error() + "\n also failed delete:" + deleteErr.Error())
```

PR #62, opened 2020-02-05
restart on env changes

jkminneti:
rollout restart on `env` commands

### DIFF
--- a/cmd/env.go
+++ b/cmd/env.go
@@ -31,7 +31,11 @@ var fileCmd = &cobra.Command{
 	Use:          "file [app] [local filepath]",
 	Short:        "batch env set",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return k8s.CreateEnvFromFile(args[0], args[1])
+		err := k8s.CreateEnvFromFile(args[0], args[1])
+		if err != nil {
+			return err
+		}
+		return k8s.Restart("deployments", args[0])
 	},
 }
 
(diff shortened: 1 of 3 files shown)


cmd/env.go
```diff
@@ -31,7 +31,11 @@ var fileCmd = &cobra.Command{
 	Use:          "file [app] [local filepath]",
 	Short:        "batch env set",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return k8s.CreateEnvFromFile(args[0], args[1])
```
Hbbb: I know we're inconsistent about this throughout the code, but maybe this is a good time to draw a line in the sand. Do we use naked returns or not? You're using them in the guts of `k8s` but not here. Sup wit dat?

Hbbb: From [A Tour of Go](https://tour.golang.org/basics/7) 

> Naked return statements should be used only in short functions, as with the example shown here. They can harm readability in longer functions.

jkminneti: Yeah I've been refactoring us out of naked returns. 

When `err` in a new scope shadows the `err` you're returning, it's an easy-to-miss problem.

We went with naked returns to avoid things like `return []byte{}, err`, but it seems like that can be refactored away without them

jkminneti: but as said in your reference, short functions should be fine. 

if there's only 1 potential error, it really is a lot cleaner

quinn: i think we should do naked returns, unless we can do this 🤓

PR #64, opened 2020-02-08
Heck pkger

jkminneti:
as promised


pkg/containers/repository.go
```diff
@@ -198,7 +197,7 @@ func convertResponse(response *http.Response) (prereleaseYamls []string, release
 			return
 		}
 
-		if !strings.HasPrefix(header.Name, core.Dir) {
+		if !strings.HasPrefix(header.Name, ".tuber") {
```
jkminneti: this stuff is from reverting the circular dependency

pkg/core/init.go
```diff
@@ -31,7 +32,7 @@ func InitTuberApp(appName string, routePrefix string) (err error) {
 }
 
 func createTuberDirectory() (err error) {
-	if err = os.Mkdir(Dir, os.ModePerm); os.IsExist(err) {
+	if err = os.Mkdir(tuberConfigPath, os.ModePerm); os.IsExist(err) {
```
jkminneti: also circular dependency related

jkminneti:
😭

PR #72, opened 2020-02-12
create namespaces with sidecar injection

jkminneti:
yes, that's all this actually does. apart from that it's refactors left and right


data/tuberapps/gen.go
```diff
@@ -31,7 +31,7 @@ func Yamls() {
 	}
 
 	for _, yaml := range yamls {
-		file, err := ioutil.ReadFile(yaml.Name())
+		file, err := ioutil.ReadFile("yamls/" + yaml.Name())
```
Hbbb: I wonder if it's preferable to use `path.Join` and declare the path & filename as locals?

data/tuberapps/rolebinding.go
```diff
@@ -0,0 +1,8 @@
+// Package data is generated
+package data
+
+// Rolebinding is generated. Returns the default rolebinding for a new tuber app
+var Rolebinding = TuberYaml{
+	Filename: "rolebinding.yaml",
+	Contents: []byte{ 107,105,110,100,58,32,82,111,108,101,66,105,110,100,105,110,103,10,97,112,105,86,101,114,115,105,111,110,58,32,114,98,97,99,46,97,117,116,104,111,114,105,122,97,116,105,111,110,46,107,56,115,46,105,111,47,118,49,98,101,116,97,49,10,109,101,116,97,100,97,116,97,58,10,32,32,110,97,109,101,58,32,116,117,98,101,114,45,97,100,109,105,110,10,32,32,110,97,109,101,115,112,97,99,101,58,32,123,123,46,110,97,109,101,115,112,97,99,101,125,125,10,114,111,108,101,82,101,102,58,10,32,32,97,112,105,71,114,111,117,112,58,32,114,98,97,99,46,97,117,116,104,111,114,105,122,97,116,105,111,110,46,107,56,115,46,105,111,10,32,32,107,105,110,100,58,32,82,111,108,101,10,32,32,110,97,109,101,58,32,116,117,98,101,114,45,97,100,109,105,110,10,115,117,98,106,101,99,116,115,58,10,32,32,45,32,107,105,110,100,58,32,83,101,114,118,105,99,101,65,99,99,111,117,110,116,10,32,32,32,32,110,97,109,101,58,32,100,101,102,97,117,108,116,10,32,32,32,32,110,97,109,101,115,112,97,99,101,58,32,116,117,98,101,114, },
```
quinn: I would prefer hex over decimal .. https://golang.org/pkg/encoding/hex/

data/tuberapps/yamls/namespace.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: {{.namespace}}
+  labels:
+    istio-injection: enabled
```
quinn: image

PR #77, opened 2020-02-19
Add apps command family

### LINKED ISSUE #75: Tuber `apps` command
- Add (adds app to the `tuber-apps` config map, doesn't do anything else)
- Remove
- Create
- List

Jefferson-Faseler:
# Resolves
#75 

# Moves
- `tuber create` -> `tuber apps create`
- `tuber list` -> `tuber apps list`

# Adds
- `tuber apps add` - which only adds an app to the tuber-config, no resource creation
- `tuber apps remove` - the inverse of `tuber apps add`, no resource destruction
- `tuber apps destroy` - removes everything in a given namespace, namespace included


cmd/appsAdd.go
```diff
@@ -0,0 +1,26 @@
...
+	SilenceUsage: true,
+	Use:          "add [app name] [docker repo] [deploy tag]",
+	Short:        "add an app to the tuber-apps config map in the current the cluster",
+	Args:         cobra.ExactArgs(3),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		appName := args[0]
+		repo := args[1]
+		tag := args[2]
+
+		return core.AddAppConfig(appName, repo, tag)
```
jkminneti: I think this one will only lead to confusion. There's no reason to add to `tuber-apps` without tuber having access to the tuberapp

cmd/appsRemove.go
```diff
@@ -0,0 +1,24 @@
...
+// appsRemoveCmd represents the appsRemove command
+var appsRemoveCmd = &cobra.Command{
+	SilenceUsage: true,
+	Use:          "remove [app name]",
+	Short:        "remove an app from the tuber-apps config map in the current cluster",
+	Args:         cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		appName := args[0]
+
+		return core.RemoveAppConfig(appName)
```
jkminneti: Im guessing the reasoning behind `add` is to add back an app after temporarily disabling CD (by running this). But - I'd rather just rerun `create` (as it should be idempotent) rather than having an extra command. Though that does make me question the naming.

jkminneti: If you agree, we should definitely keep Remove though - we're using tuber's role for pod-wide rolebindings

Jefferson-Faseler: It either can't be named `create` or we allow the cd to stop by removing it from tuber with an inverse command to simply add it back.

I just used this to switch release cutter from the tuberfy tag to the master tag. It was easy and safe.

I think we could rename the command, but ultimately it is good to have IMO. Re-running create is much less intuitive to me.

pkg/core/destroy_tuber_app.go
```diff
@@ -0,0 +1,16 @@
...
+
+import "tuber/pkg/k8s"
+
+// DestroyTuberApp deletes all resources for the given app on the current cluster
+func DestroyTuberApp(appName string) (err error) {
+	if err = k8s.Delete("namespace", appName, appName); err != nil {
+		return
+	}
+
+	if err = RemoveAppConfig(appName); err != nil {
```
Jefferson-Faseler: Is this just unnecessary now? Could do `return RemoveAppConfig` instead?

Jefferson-Faseler: @jkminneti

jkminneti: Yup - of course either one works, whatever's clearer. I usually like the `return whatever()` cus it keeps it from looking like another early break, but then again it's a bit less work to add to the function if all of them are `if`s

Jefferson-Faseler: Kind of what I was thinking, either way this seems to go against the style of the repo ATM. I'll switch it.

jkminneti: Yeah it does. We should sorta-kinda stop with the naked returns also but :shrug:. I want to go by the style of stlib code, not the existing code

Jefferson-Faseler: We were all about naked returns a month ago

jkminneti: 1. Write function with naked returns
2. Add `for` loop to function
3. Have explicit returns in the loop, naked returns everywhere else, or refactor it all

Got frustrated with how often that was happening, so I'd rather have explicit returns the whole way down. Even though the magic's nice

jkminneti: 1. Write function with naked returns
2. Add `for` loop to function
3. Have explicit returns in the loop, naked returns everywhere else, or refactor it all

Got frustrated with how often that was happening, so I'd rather have explicit returns the whole way down. Even though the magic's nice

jkminneti: We liked em because `return someStruct{}, err` sucks. Buuut just pass pointers then - `return nil, err` is way better all around. Or declare in a local `var hi someStruct`, and _set_ it's value in the success case (which is how naked returns work anyway)

cmd/appsRemove.go
jkminneti: I'm worried someone would remove and re-add to rename the app, or just not realize create is a thing

jkminneti: In both cases it'd be hard to tell why it's broken, and in both cases it would look successful to the client - the error wouldn't show up until tuber tried to deploy the app from an incoming build

Jefferson-Faseler: Right, but remove and destroy are two different commands. So that would be valid.

It's almost as if we want `apply` oor `new` instead of `create` and `watch` instead of `add`.

Eventual idempotency of create sounds appealing, but that's also tricky since it isn't applying it is creating a namespace and configs.

Jefferson-Faseler: What if we borrowed from Helm's verbage? `tuber apps install` instead of `create`?

jkminneti: I like that. And I'll get it idempotent tomorrow, shouldn't be too bad. Basically just skip the loop if the namespace exists

cmd/appsInstall.go
```diff
@@ -0,0 +1,26 @@
...
+	SilenceUsage: true,
+	Use:          "install [app name] [docker repo] [deploy tag]",
+	Short:        "install a new app in the current cluster",
+	Args:         cobra.ExactArgs(3),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		appName := args[0]
+		repo := args[1]
+		tag := args[2]
+
+		return core.InstallTuberApp(appName, repo, tag)
```
jkminneti: 🌜 undefined

pkg/core/destroy_tuber_app.go
```diff
@@ -0,0 +1,12 @@
+package core
```
Hbbb: Can we create `pkg/core/apps.go` or `pkg/core/tuber_apps.go` that these functions live in. Regardless of the name, this and `CreateTuberApp` belong in the same file.

PR #78, opened 2020-02-19
Make app name a flag

Hbbb:


cmd/create.go
```diff
@@ -9,11 +9,10 @@ import (
 // createCmd represents the create command
 var createCmd = &cobra.Command{
 	SilenceUsage: true,
-	Use:          "create [app name] [docker repo] [deploy tag]",
+	Use:          "create [docker repo] [deploy tag]",
 	Short:        "create new app in current cluster",
-	Args:         cobra.ExactArgs(3),
+	Args:         cobra.ExactArgs(2),
 	RunE: func(cmd *cobra.Command, args []string) error {
-		appName := args[0]
```
jkminneti: this one feels weird - feels like `-a` indicates an existing app in all other places - idk

cmd/deploy.go
```diff
@@ -12,10 +12,9 @@ import (
 
 var deployCmd = &cobra.Command{
 	SilenceUsage: true,
-	Use:          "deploy [appName]",
+	Use:          "deploy",
 	Short:        "Deploys an app",
 	RunE:         deploy,
-	Args:         cobra.ExactArgs(1),
```
jkminneti: I think we'll want to make it required on each of these:
https://github.com/spf13/cobra#required-flags

PR #87, opened 2020-02-26
Create install.sh

quinn:
to test:

```
curl https://raw.githubusercontent.com/Freshly/tuber/install-script/install.sh | bash
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -4,16 +4,18 @@ tmpfile=$(mktemp)
 installfile=/usr/local/bin/tubectl
 os=$(uname)
 
+echo
+
 case $os in
   Darwin)
     echo "Installing for OS X..."
-    curl -Lo $tmpfile https://github.com/Freshly/tuber/releases/download/v1.0/tuber_macos
+    curl -Lo $tmpfile https://github.com/Freshly/tuber/releases/download/v1.0/tuber_macos &> /dev/null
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+tmpfile=$(mktemp)
+installfile=/usr/local/bin/tubectl
+
+curl -Lo $tmpfile https://github.com/Freshly/tuber/releases/download/v1.0/tuber_macos
```
Hbbb: This should read the OS type and install the correct binary based on that 💪

quinn: @Hbbb 💀 ok fine

install.sh
```diff
@@ -0,0 +1,26 @@
...
+case $os in
+  Darwin)
+    echo "Installing for OS X..."
+    curl -Lo $tmpfile https://github.com/Freshly/tuber/releases/download/v1.0/tuber_macos &> /dev/null
+    mv $tmpfile $installfile
+    chmod +x $installfile
+    ;;
+  Linux)
+    echo "Installing for Linux..."
+    curl -Lo $tmpfile https://github.com/Freshly/tuber/releases/download/v1.0/tuber_macos &> /dev/null
```
jkminneti: `tuber_macos`

Issue #102, opened 2020-03-24
tuber cant create itself with tuber commands


jkminneti:
for reasons

PR #104, opened 2020-03-24
plant

jkminneti:
🌱 
resolves https://github.com/Freshly/tuber/issues/102


data/tuberapps/namespace.go
```diff
@@ -4,5 +4,5 @@ package data
 // Namespace is generated. Returns the default namespace for a new tuber app
 var Namespace = TuberYaml{
 	Filename: "namespace.yaml",
-	Contents: []byte{ 97,112,105,86,101,114,115,105,111,110,58,32,118,49,10,107,105,110,100,58,32,78,97,109,101,115,112,97,99,101,10,109,101,116,97,100,97,116,97,58,10,32,32,110,97,109,101,58,32,123,123,46,110,97,109,101,115,112,97,99,101,125,125,10,32,32,108,97,98,101,108,115,58,10,32,32,32,32,105,115,116,105,111,45,105,110,106,101,99,116,105,111,110,58,32,101,110,97,98,108,101,100, },
+	Contents: []byte{ 97,112,105,86,101,114,115,105,111,110,58,32,118,49,13,10,107,105,110,100,58,32,78,97,109,101,115,112,97,99,101,13,10,109,101,116,97,100,97,116,97,58,13,10,32,32,110,97,109,101,58,32,123,123,46,110,97,109,101,115,112,97,99,101,125,125,13,10,32,32,108,97,98,101,108,115,58,13,10,32,32,32,32,105,115,116,105,111,45,105,110,106,101,99,116,105,111,110,58,32,123,123,46,105,115,116,105,111,69,110,97,98,108,101,100,125,125, },
```
jkminneti: reminder to myself: regen on mac, i think windows is messing with it

cmd/apps.go
```diff
@@ -13,17 +14,32 @@ var appsCmd = &cobra.Command{
 	Short: "A root command for app configurating.",
 }
 
+var istioEnabled bool
+
 var appsInstallCmd = &cobra.Command{
 	SilenceUsage: true,
-	Use:          "install [app name] [docker repo] [deploy tag]",
+	Use:          "install [app name] [docker repo] [deploy tag] [--no-istio (optional flag)]",
```
Jefferson-Faseler: Maybe it's just me but I think Istio is the exception, and not the rule. Can we default the flag to false? 

If we think that is too annoying for the main audience (us, who use istio) the default for the --istio/--no-istio could be controlled by an env var, similar to the debug var, and overridden by the flag.

pkg/core/create_tuber_app.go
```diff
@@ -2,37 +2,28 @@ package core
 
 import (
 	"fmt"
-	data "tuber/data/tuberapps"
+	yamls "tuber/data/tuberapps"
```
Jefferson-Faseler: Thank you

cmd/apps.go
Jefferson-Faseler: Also, 
```
go run main.go apps install something us.gcr.io/freshly-docker/tyler_poc master --no-istio
Error: unknown flag: --no-istio
```

jkminneti: will fix, but idk. istio seems to me to be pretty core to what tuber is doing - providing full control over your app in the app's own yamls. 

there isn't really a solution aside from istio where the act of adding a fully connected app can happen entirely within its own namespace

jkminneti: will fix, but idk, I imagine the opposite, that _not_ using istio is the exception

Jefferson-Faseler: So, Tuber _wants_ you to use Istio, but if you so choose to not, that's fine too.

Jefferson-Faseler: I guess that's fine. Just trying to think about the other apps which don't make sense to have behind the gateways. Which admittedly are just edge cases against the big problem Tuber solves. Maintaining a cluster and it's continuous deployment.

jkminneti: like non-istio things are what, cardinal/birdcage/tuber - I don't see that list growing all that much, compared to normal apps

Jefferson-Faseler: Well that's only true for _our_ use case.

Regardless, I think making Istio the convention for Tuber is fine.

Issue #89, opened 2020-02-28
pull credentials from cluster


jkminneti:
🎉 jordan got his way

this would open the door to `apps install` resulting in a deployed app. or otherwise, using `deploy` like ever. 

creds are used only for gcr, k8s rolebindings would still be in play

PR #109, opened 2020-03-27
what if the binary didnt need env or creds

jkminneti:
resolves https://github.com/Freshly/tuber/issues/89

we could even kill prefixing #justsayingg


pkg/k8s/secrets.go
```diff
@@ -44,6 +44,23 @@ func CreateTuberCredentials(path string, namespace string) (err error) {
...
+	for k, v := range config.Data {
+		decoded, decodeErr := base64.StdEncoding.DecodeString(v)
+		if decodeErr != nil {
+			return nil, decodeErr
+		}
+		config.Data[k] = string(decoded)
+	}
+
+	return config, nil
+}
```
quinn: How did we not have this already? should some things be refactored to depend on this?

jkminneti: First time we've needed to read a secret, aside from env list

quinn: thats my point. seems like we should have one function to read a secret

Issue #99, opened 2020-03-23
Add confirmation step to write commands


quinn:
any command that alters anything on the cluster should have a confirmation.
any command that reads data and does not alter anything should not require confirmation. 

print context name to STDERR.

PR #111, opened 2020-04-01
prompt or display cluster on commands

jkminneti:
resolves https://github.com/Freshly/tuber/issues/99


cmd/helpers.go
```diff
@@ -20,3 +23,43 @@ func credentials() (creds []byte, err error) {
...
+
+	if err != nil {
+		if err.Error() == "unexpected newline" {
+			return nil
+		} else if err.Error() == "EOF" {
+			return fmt.Errorf("cancelled")
+		} else {
+			return err
+		}
+	}
```
jkminneti, line 54: yes, `fmt.Scanln` sucks. open to any suggestions for how to scan until newline without _erroring if it's only a newline_ would be appreciated

PR #112, opened 2020-04-01
tuberrc

jkminneti:
```
clusters:
  staging: gke_crazy_staging_cluster_name
  production: gke_crazy_production_cluster_name
```
```
tuber config
tuber switch staging
```


cmd/helpers.go
```diff
@@ -20,3 +21,27 @@ func credentials() (creds []byte, err error) {
 	creds, err = ioutil.ReadFile(credentialsPath)
 	return
 }
+
+func getTuberrc() (*tuberrc, error) {
+	path := viper.GetString("tuberrc-path")
```
quinn, line 26: i think this should default to `~/.config/tuber`, and should be created automatically if it doesn't exist.

jkminneti:
works on windows and mac - doesn't work on WSL. It might work on normal linux though. Either way, let's call it a bugfix for later - try this and you'll see why

cmd/switch.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	var clusterName string
+	var displayCluster string
+
+	clusterName = args[0]
+	displayCluster = clusterName
+
+	if c != nil {
+		resolvedShorthand := c.Clusters[args[0]]
```
Hbbb, line 31: You made `clusterName` on line 27, you should use that here.

PR #119, opened 2020-04-07
Revert "plant"

jkminneti:
Reverts Freshly/tuber#104


quinn:
I don't think this is necessary anymore.

PR #129, opened 2020-08-06
Support Multiple Pending Releases

Hbbb:
Currently, Tuber finds pending releases by matching on image tag. This doesn't work for apps like the resque dashboard, which are deployed as three instances of the same app, with the same image tag, and different names. 

For example, `resque-admin-prod`, `resque-admin-staging`. When running `tuber deploy`, tuber would find the first matching image tag and deploy that app. 

This change allows tuber to find all apps with a matching image tag and deploy them.

### DIFF
--- a/pkg/events/filter.go
+++ b/pkg/events/filter.go
@@ -5,17 +5,19 @@ import (
 	"tuber/pkg/listener"
 )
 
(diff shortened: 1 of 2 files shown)


jkminneti:
done in the reviewapps stuff (streamer refactor)

PR #130, opened 2020-08-27
I think this is all I need to do?

Jefferson-Faseler:
Right now, when there is a config error or something invalid about applying the tubernetes for an event that event will be re-returned to the pubsub queue because the deadline was extended, but never ack'd. What ends up happening is a simple typo plugs up the pubsub queue for a tuber and everything needs to get purged or the specific event needs to be surgically removed.

Solution: Ack an event if there was an error, assuming any error will always re-fail. This probably isn't true about something like a tuber permission or config var, but that scenario is much easier to resolve with a `tuber deploy` than clearing pubsub messages.


jkminneti:
Done in the review apps stuff, ty for the investigation and sorry to steal the credit lolll. ended up redoing the whole streamer

PR #131, opened 2020-08-29
Review Apps

jkminneti:
1. 👉 duplicate resources
2. #132 trigger builds and stuff
3. #133 filter for review apps

testing:
```go
package cmd

import (
	"tuber/pkg/core"

	"github.com/spf13/cobra"
)

var dupCmd = &cobra.Command{
	SilenceUsage: true,
	Use:          "dup [source app] [target app]",
	Short:        "dup dup",
	Args:         cobra.ExactArgs(2),
	RunE:         dup,
}

func dup(_ *cobra.Command, args []string) error {
	return core.NewReviewAppSetup(args[0], args[1])
}

func init() {
	rootCmd.AddCommand(dupCmd)
}
```

includes overdue `go mod tidy`, don't @ me


pkg/core/review_apps.go
```diff
@@ -0,0 +1,109 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+	"tuber/pkg/k8s"
+)
+
+// NewReviewAppSetup replicates a namespace and its roles, rolebindings, and secrets after removing their non-generic metadata.
+// Also renames source app name matches across all relevant resources.
+func NewReviewAppSetup(sourceApp string, targetApp string) error {
```
Hbbb, line 12: Personal preference, but I prefer grouping arguments with the same type like `func NewReviewApp(sourceApp, targetApp string)`.

pkg/core/review_apps.go
```diff
@@ -0,0 +1,109 @@
+package core
```
Hbbb, line 1: What do you think about renaming `targetApp` to `reviewApp`. To me, that makes it easier to follow the code.

PR #133, opened 2020-09-02
filter for review apps

jkminneti:
1. #131 duplicate resources
2. #132 trigger builds and stuff
3. 👉 filter for review apps

also flips the whole stupid streamer on its head


pkg/core/tuber.go
```diff
@@ -105,7 +109,7 @@ func TuberApps() (apps AppList, err error) {
 	mutex.Lock()
 	defer mutex.Unlock()
 	if cache == nil || cache.isExpired() {
-		apps, err = getTuberApps()
```
Hbbb, line 108: Maybe we should rename `TuberApps` to `SourceApps`. We can probably drop the word Tuber from the other function name too. So we end up with `SourceApps`, `ReviewApps`, and `SourceAndReviewApps`.

cmd/start.go
```diff
@@ -96,9 +96,17 @@ func start(cmd *cobra.Command, args []string) {
...
-	go streamer.Stream(unprocessedEvents, processedEvents, failedEvents, errReports)
+	eventProcessor := events.EventProcessor{
+		Creds:             creds,
+		Logger:            logger,
+		ClusterData:       data,
+		ReviewAppsEnabled: viper.GetBool("reviewapps-enabled"),
+		Unprocessed:       unprocessedEvents,
+		Processed:         processedEvents,
+		ChErr:             failedEvents,
+		ChErrReports:      errReports,
```
Hbbb, line 107: I like how in the `deploy` command, you renamed `errReports` into `sentryErrors`. Can we do that here too with the var, and to the struct property `ChErrReports` too? It seems like Sentry is here to stay and naming this explicitly makes the code more comprehensible.

pkg/events/processor.go
```diff
@@ -0,0 +1,104 @@
...
+			}
+
+			p.processEvent(event, apps)
+		}(event)
+	}
+	wait.Wait()
+}
+
+func (p EventProcessor) apps() ([]core.TuberApp, error) {
+	if p.ReviewAppsEnabled {
```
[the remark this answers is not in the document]
jkminneti: Oops yes lol

cmd/start.go
jkminneti: Soo the only reason I did that is because I consider `deploy` as temporary code. I still think sentry is an integration, and that an `errors` package should pass things to a configurable integration to use (in our case sentry).

I think the better names for these channels are `Failures` or `FailedReleases`, and `Errors`?

pkg/core/tuber.go
jkminneti: Totally down

PR #134, opened 2020-09-02
POPS-512 GRPC Client & Server

Hbbb:


cmd/reviewApps.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/spf13/cobra"
+)
+
+// reviewAppsCmd represents the reviewApps command
+var reviewAppsCmd = &cobra.Command{
+	Use:   "review-apps",
+	Short: "A brief description of your command",
+	Long:  "",
+	Run: func(cmd *cobra.Command, args []string) {
+		fmt.Println("reviewApps called")
```
jkminneti, line 35: doesn't need a `Run`

pkg/client/client.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"log"
+	"tuber/pkg/proto"
+
+	"google.golang.org/grpc"
+)
+
+func NewClient() (proto.TuberServiceClient, *grpc.ClientConn) {
+	var conn *grpc.ClientConn
+	conn, err := grpc.Dial(":9000", grpc.WithInsecure())
```
jkminneti, line 12: can we hook this up to the url? i think we can test the whole stack today

also since that's probably an env var, the url will need to become a param

pkg/k8s/kubectl.go
```diff
@@ -142,5 +144,73 @@ func CurrentCluster() (string, error) {
...
+		UserData userData `yaml:"user"`
+	}
+
+	type clusterUsers struct {
+		Users []clusterUser `yaml:"users"`
+	}
+
+	var tmp clusterUsers
+
+	out, err := kubectl([]string{"config", "view", "--raw"}...)
```
jkminneti, line 177: imo, this is the only part that should live in `kubectl`, and _maaaybe_ the only part that should live in `k8s`

we also might have other uses for a parsed local config, so this struct should maybe exported and available through a separate function

also just noting that this format (everything under `auth-provider`) is GCP-specific. I think tuber is very much a GCP tool so it's fine, but just throwing it out there

pkg/k8s/kubectl.go
```diff
@@ -142,5 +144,73 @@ func CurrentCluster() (string, error) {
...
+	return authToken, nil
+}
+
+// CanDeploy determines if the current user can create a deployment
+func CanDeploy(appName, token string) bool {
+	t := fmt.Sprintf("--token=%s", token)
+
+	out, err := kubectl([]string{"auth", "can-i", "create", "deployments", "-n", appName, t}...)
+	if err != nil {
+		return false
```
jkminneti, line 210: output on this could be useful - i think this should maybe return an error

pkg/server/server.go
```diff
@@ -0,0 +1,66 @@
...
+	"tuber/pkg/k8s"
+	"tuber/pkg/proto"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/reflection"
+)
+
+// Serve serves
+func Serve() {
+	lis, err := net.Listen("tcp", ":9000")
```
jkminneti, line 16: i think our convention is to listen on 3000, or at least that's what we've told app devs

pkg/server/server.go
```diff
@@ -0,0 +1,66 @@
...
+
+	if err != nil {
+		log.Fatalf("failed to listen: %v", err)
```
jkminneti, line 19: this func doesnt block until `grpc.Serve`, so this should probably return an error and panic out in `start.go`

pkg/server/server.go
```diff
@@ -0,0 +1,66 @@
...
+type Server struct {
+}
+
+// Authenticate authenticates
+func (s *Server) Authenticate(appName, token string) bool {
+	return k8s.CanDeploy(appName, token)
+}
+
+// CreateReviewApp creates a review app
+func (s *Server) CreateReviewApp(ctx context.Context, req *proto.Request) (*proto.Response, error) {
```
jkminneti, line 44: does this give us a specific url out of the box? or is it listening on `/`?

also this should probably be its own file, we already have an idea for another endpoint (to replace the deploy command)

pkg/server/server.go
```diff
@@ -0,0 +1,66 @@
...
+	var err error
```
jkminneti, line 45: what's this for?

cmd/start.go
```diff
@@ -55,6 +56,7 @@ func start(cmd *cobra.Command, args []string) {
 	defer close(errReports)
 
 	go sentry.Stream(sentryEnabled, sentryDsn, errReports, logger)
+	go server.Serve()
```
jkminneti, line 59: could be useful for debugging to have a separate command that _just_ does this... not it should be committed to code or not though

cmd/reviewApps.go
```diff
@@ -0,0 +1,89 @@
...
+	},
+}
+
+var reviewAppsCreateCmd = &cobra.Command{
+	Use:   "create",
+	Short: "A brief description of your command",
+	Long:  "",
+	Args:  cobra.ExactArgs(2),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		appName := args[0]
```
jkminneti, line 45: should probably be specific to _source_ app name

pkg/client/client.go
jkminneti: oh right - nevermind on the env var. `tuber config` needs to become:
```
clusters:
  staging: 
    name: asdf
    url: asdf
  production: 
    name: asdf
    url: asdf
  internal: 
    name: asdf
    url: asdf
```

i can hook that up quick, the config code is pretty weird

PR #136, opened 2020-09-03
make sentry an integration

jkminneti:
no, it's not necessary. but something like this would make me think of this as open-source so you'd have that off your back


quinn-freshly:
@jkminneti i think all of this is already in now that we've merged review apps. is there anything in here we still need?

jkminneti:
this is something else entirely, i can explain in stand - also have a massive conflict to resolve looks like

PR #138, opened 2020-09-08
POPS-512 GRPC Client & Server - Create Review Apps

Hbbb:
https://freshly.atlassian.net/browse/POPS-512


pkg/reviewapps/client.go
```diff
@@ -0,0 +1,22 @@
...
+
+	"google.golang.org/grpc"
+)
+
+// NewClient returns a GRPC client
+func NewClient(url string) (proto.TuberClient, *grpc.ClientConn) {
+	// fullURL := fmt.Sprintf("%s:9000", url)
+
+	var conn *grpc.ClientConn
+	conn, err := grpc.Dial(":9000", grpc.WithInsecure())
```
Hbbb, line 15: This needs the full URL passed in

pkg/reviewapps/permissions.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"tuber/pkg/core"
+	"tuber/pkg/k8s"
+
+	"github.com/spf13/viper"
+)
+
+func canCreate(appName, token string) bool {
+	return appName != "tuber" &&
+		viper.GetBool("review-apps") &&
```
Hbbb, line 12: This should be passed into the server upon start up? Maybe?

pkg/reviewapps/server.go
```diff
@@ -0,0 +1,64 @@
...
+			}, nil
+		}
+
+		return &proto.CreateReviewAppResponse{
+			Error: err.Error(),
+		}, nil
+	}
+
+	// TODO: Where can I get the project name?
+	removeTrigger, err := CreateAndRunTrigger(ctx, []byte(""), in.AppName, "<PROJECT>", reviewAppName, in.Branch)
```
Hbbb, line 42: Project name should be passed in from `cmd` I assume.

Hbbb: Also, this pattern doesn't make sense. The only time the `removeTrigger` function is returned is when there is no error. If `err != nil`, `removeTrigger` will always be nil.

pkg/server/grpc.go
```diff
@@ -0,0 +1,30 @@
...
+	"net"
+	"tuber/pkg/proto"
+	"tuber/pkg/reviewapps"
+
+	"google.golang.org/grpc"
+)
+
+// Start starts a GRPC server
+func Start() error {
+	lis, err := net.Listen("tcp", ":9000")
```
Hbbb, line 14: This shouldn't be hardcoded. Where should it come from?

pkg/server/grpc.go
```diff
@@ -0,0 +1,30 @@
...
+	if err != nil {
+		return err
+	}
+
+	s := reviewapps.Server{}
+
+	server := grpc.NewServer()
+	proto.RegisterTuberServer(server, &s)
+
+	fmt.Println("starting GRPC server")
```
Hbbb, line 24: This needs the actual logger.

pkg/proto/tuber_service.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
+package proto;
```
Hbbb, line 2: Probably change the package name for this

PR #140, opened 2020-09-14
Improve Logging

Hbbb:
This is mostly for review apps, and for removing debug logging.


jkminneti:
could switch these over to `Debug`'s

PR #142, opened 2020-09-17
POPS-539 Use {{.tuberAppName}} in YAMLs in init command

Hbbb:


pkg/core/init.go
```diff
@@ -46,16 +46,13 @@ func createTuberDirectory() error {
 }
 
 func createDeploymentYAML(appName string) error {
-	templateData := map[string]string{
-		"appName": appName,
-	}
+	templateData := map[string]string{}
```
jkminneti, line 49: what's this stuff about? seems like this wouldnt work

jkminneti (changes requested):
`go generate ./...`
`data\tuberapps\gen.go` for context

quinn-freshly (approved):

pkg/core/init.go
Hbbb: It's because the underlying function takes in a `map[string]string` and I didn't want to change it to handle nil. This works.

jkminneti (approved):

PR #145, opened 2020-10-01
Fix Tuber! (and add debug logging)

Jefferson-Faseler:
Adds various debug logs and a couple useful info logs. This should help anyone struggling to see why their code is not being deployed. I was tempted to do some refactoring but I'll just stick to these simple additions for now.

## What was the bug?
Tuber wasn't writing to the `processed` events channel for non-deploy events. That meant the logger never ack'd and read messages off the channel, blocking the app.


pkg/events/processor.go
```diff
@@ -50,21 +55,28 @@ func (p EventProcessor) Start() {
...
+		zap.String("digest", event.Digest),
+	)
+	p.Logger.Debug("Listing tuber apps",
+		zap.Any("apps", apps),
+	)
 
 	for _, app := range apps {
 		if app.ImageTag == event.Tag {
 			p.runDeploy(app, event)
+		} else {
```
Hbbb, line 78: This is the bug fix

jkminneti (approved):

Hbbb (approved):

PR #146, opened 2020-10-01
Stop closing closed channels

Hbbb:


Jefferson-Faseler (changes requested):
I think it is probably better if we defer to closing this in the processor. We want the errors channel open at the highest level I imagine.

jkminneti (approved):

Jefferson-Faseler (approved):

PR #149, opened 2020-10-02
Add portforwarding command

Jefferson-Faseler:
Adds `port-forward` command. Can be used with all features of the `kubectl` command of the same name.
`tuber port-forward -a user-service 8989`
To open a port on localhost:8989 -> pod::8989

Ports can be a variadic list that also take advantage of the `:` syntax in the `kubectl port-forward` [command](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#port-forward). 

`tuber port-forward -a user-service -p user-service-8822272-fsdalhj 9001` to specify a particular pod.

Able to specify the workload, local IP addresses, and pod wait time in addition to the small examples above.

Also, as an added bonus I gave `exec` the ability to specify a pod as well. Specifying a pod has many added benefits, like faster command times and edge-cases for a-typically configured pods.


cmd/helpers.go
```diff
@@ -13,6 +14,12 @@ import (
 	"github.com/spf13/viper"
 )
 
+var address string
+var appName string
+var pod string
+var podRunningTimeout string
+var workload string
```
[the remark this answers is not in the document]
jaysonesmith: Oh wait, are these being set by Cobra?

jaysonesmith (commented):
Not sure my reviews are necessarily going to be super useful right now, but I've asked some questions to try to get a better understanding and stuff.

cmd/helpers.go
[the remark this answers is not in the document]
Jefferson-Faseler: Yes, they are set by Cobra in commands that share the need for these vars; `exec`, and `port-forward`.

cmd/portForward.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	ports := args
+
+	portForwardArgs := []string{"--address", address, "--pod-running-timeout", podRunningTimeout}
+	err = k8s.PortForward(podName, appName, ports, portForwardArgs...)
+	return err
+}
+
+func init() {
+	portForwardCmd.Flags().StringVar(&address, "address", "localhost", "specify an address on your local machine to forward from. Can be a comma separated list. Only IP addresses and 'localhost' are valid.")
```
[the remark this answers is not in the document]
Jefferson-Faseler: Yeah, let k8s do the work for parsing it. No need to reinvent the wheel here IMO.

cmd/exec.go
```diff
@@ -15,50 +13,29 @@ var execCmd = &cobra.Command{
...
-
-	labels := strings.TrimSuffix(string(l), ",")
-
-	jsonPath := fmt.Sprintf(`-o=jsonpath="%s"`, `{.items[0].metadata.name}`)
-	name, err := k8s.GetCollection("pods", appName, "-l", labels, jsonPath)
-	if err != nil {
-		return err
+		containerName = fetchWorkload()
 	}
+	podName, err := fetchPodname()
```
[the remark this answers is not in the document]
Jefferson-Faseler: Oh gosh. 🤦 Nice catch

cmd/exec.go
```diff
@@ -15,50 +13,29 @@ var execCmd = &cobra.Command{
...
-		return err
-	}
-
-	labels := strings.TrimSuffix(string(l), ",")
-
-	jsonPath := fmt.Sprintf(`-o=jsonpath="%s"`, `{.items[0].metadata.name}`)
-	name, err := k8s.GetCollection("pods", appName, "-l", labels, jsonPath)
-	if err != nil {
-		return err
+		containerName = fetchWorkload()
```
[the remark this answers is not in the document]
Jefferson-Faseler: It is also used in https://github.com/Freshly/tuber/blob/add-portforwarding-command/cmd/helpers.go#L196 for fetching pod names.

jkminneti (changes requested):
using the simplest example, 
```
tuber exec -a user-service -w user-service-sidekiq
```
it returns 
```
error: you must specify at least one command for the container
```
upon the kubectl exec.

debug logs indicate the pod name was successfully identified.

Jefferson-Faseler:
> using the simplest example,
> 
> ```
> tuber exec -a user-service -w user-service-sidekiq
> ```
> 
> it returns
> 
> ```
> error: you must specify at least one command for the container
> ```
> 
> upon the kubectl exec.
> 
> debug logs indicate the pod name was successfully identified.

I'm confused. Are you suggesting I make changes to the `exec` command? I think the output would describe why you are seeing the error.

cmd/portForward.go
```diff
@@ -0,0 +1,53 @@
...
+	Long: `Forward requests from a local address and port to a running pod.
+You are able to specify multiple addresses and ports, but all combinations must be valid and running.
+
+This command will always run against a single pod until either that pod terminates or this command is closed.
+
+Specifying a workload:
+The workload name will default to the app name if not supplied. If the workload name is not the same as the app name, that argument will be required to run the command successfully.
+
+For example. When the desired workload is a deployment named 'user-service-sidekiq' within a tuber app named 'user-service':
+tuber exec -a user-service -w user-service-sidekiq
```
jkminneti, line 24: not exec

Jefferson-Faseler: Oh hey, good catch

jkminneti: i had copy pasted from that line in the docs (to explain my previous comment lol)

cmd/portForward.go
```diff
@@ -0,0 +1,53 @@
...
+	"tuber/pkg/k8s"
+
+	"github.com/spf13/cobra"
+)
+
+// portForwardCmd represents the portForward command
+var portForwardCmd = &cobra.Command{
+	SilenceUsage: true,
+	Use:          "port-forward -a appName [port(s)]",
+	Args:         cobra.MinimumNArgs(1),
```
jkminneti, line 13: exactargs?

Jefferson-Faseler: Nope, as many as your ❤️ desires

Jefferson-Faseler: Well, technically 65535 args max

jkminneti: why though? not like it's variadic

Jefferson-Faseler: It is https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#port-forward

jkminneti: `one or more local ports to a pod` why would anyone... but whatever you're absolutely right

cmd/portForward.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"tuber/pkg/k8s"
+
+	"github.com/spf13/cobra"
+)
+
+// portForwardCmd represents the portForward command
+var portForwardCmd = &cobra.Command{
+	SilenceUsage: true,
+	Use:          "port-forward -a appName [port(s)]",
```
jkminneti, line 12: ok call it pedantic, but i don't like the `-a`. could easily do `tuber port-forward payment-service 8080`

in fact, what if we assumed 8080 by default

cmd/portForward.go
```diff
@@ -0,0 +1,53 @@
...
+func portForward(cmd *cobra.Command, args []string) error {
+	podName, err := fetchPodname()
+	if err != nil {
+		return err
+	}
+	ports := args
+
+	portForwardArgs := []string{"--address", address, "--pod-running-timeout", podRunningTimeout}
+	err = k8s.PortForward(podName, appName, ports, portForwardArgs...)
+	return err
```
jkminneti, line 42: this should at least print a clickable link. bonus brownie taco pizza points if it were to open your default web browser to it

Jefferson-Faseler: I don't think I'm going to do this change. A brownie taco pizza sounds very unpleasant.

cmd/portForward.go
Jefferson-Faseler: I'm matching what exec's syntax is so ditching `-a` would change that command as well. 

Supplying port(s) is not difficult for a user to do. I would think it is better to have tuber complain someone did not pass a port so they know it's a thing.

cmd/portForward.go
jkminneti: not even the clickable link though?

cmd/portForward.go
jkminneti: for us it makes sense, is intuitive, and is easy. but for someone who doesnt understand what port-forwarding is, some sensible defaults could get them into their sidekiq-web. 

i'm imagining "which port do i need for sidekiq" kinds of questions

cmd/portForward.go
Jefferson-Faseler: I can look into it 😒

cmd/portForward.go
Jefferson-Faseler: For sidekiq, sure. What about postgres, redis, memcached, applications (tls and http)? Then the portforwarding doesn't work by default and we have the same question

jkminneti: that's the thing, though - this isn't supposed to be a `kubectl` wrapper, it's supposed to be a "service mesh tool for app devs"

jkminneti (approved):

PR #152, opened 2020-10-14
Add filename logging to tuber init

jaysonesmith:
Quick update.

### DIFF
--- a/pkg/core/init.go
+++ b/pkg/core/init.go
@@ -1,6 +1,7 @@
 package core
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	data "tuber/data/tuberapps"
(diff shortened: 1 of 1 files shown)


jkminneti (approved):

pkg/core/init.go
```diff
@@ -88,5 +89,7 @@ func writeYAML(app data.TuberYaml, templateData map[string]string) error {
 		return err
 	}
 
+	fmt.Printf("writing %q to %q\n", app.Filename, tuberConfigPath)
```
jkminneti, line 92: actually maybe this should go _after_ a successful ioutil.WriteFile (as in "_wrote_")

jkminneti (approved):

PR #154, opened 2020-10-23
Channels are for quitters

jkminneti:
(╯°□°)╯︵ ┻━┻

- no more channels
- sentry way better
- logging better for startup and deploying


pkg/events/processor.go
```diff
@@ -1,131 +1,146 @@
...
 
-func (p EventProcessor) processEvent(event *listener.RegistryEvent, apps []core.TuberApp) {
-	p.Logger.Info("Processing event",
-		zap.String("tag", event.Tag),
-		zap.String("digest", event.Digest),
-	)
-	p.Logger.Debug("Listing tuber apps",
-		zap.Any("apps", apps),
-	)
+func (p Processor) processEvent(event *Event, apps []core.TuberApp) {
```
Hbbb, line 85: Small nitpick, I think this function is pointless. I'd rather see the logic in here moved up into `ProcessMessage`.

pkg/messages/listener.go
```diff
@@ -0,0 +1,86 @@
+package messages
```
Hbbb, line 1: Why not just call this package pubsub? We refer to it directly as pubsub in the code elsewhere already. What about something like:

```go
subscriber, err := pubsub.Subscriber(...)

err = subscriber.Listen() 
// OR
err = subscriber.Start()
```

It feels like one less mental mapping to do if we call this pubsub.

Hbbb (approved):
This looks great. I have but a few small naming suggestions.

pkg/events/processor.go
jkminneti: agreed, done

pkg/messages/listener.go
jkminneti: `pubsub` is the google library, that's why

Hbbb: Yea, you can alias that in our `pubsub` package. Our code should only ever import our `pubsub` implementation, not Google's.

jkminneti: ok did it, good suggestion 👍

jkminneti: ah oops ok nope, circular dependencies :/

cmd/start.go
```diff
@@ -42,100 +41,83 @@ func bindShutdown(logger *zap.Logger, cancel func()) {
...
-	unprocessedEvents, processedEvents, failedEvents, err := l.Listen(ctx, creds)
+	data, err := clusterData()
 	if err != nil {
+		startupLogger.Warn("failed to get cluster data", zap.Error(err))
+		report.Error(err, scope.WithContext("getting cluster data"))
 		panic(err)
 	}
 
-	data, err := clusterData()
+	listener, err := pubsub.NewListener(
```
Hbbb, line 74: This makes me so happy.

Hbbb (approved):

cmd/deploy.go
```diff
@@ -62,45 +56,15 @@ func deploy(cmd *cobra.Command, args []string) error {
...
 
-	ackable := emptyAckable{}
-	deployEvent := listener.RegistryEvent{
-		Action:  "INSERT",
-		Digest:  app.RepoHost + "/" + app.RepoPath + "@" + sha,
-		Tag:     app.ImageTag,
-		Message: ackable,
-	}
+	processor := events.NewProcessor(ctx, logger, creds, data, viper.GetBool("reviewapps-enabled"))
+	digest := app.RepoHost + "/" + app.RepoPath + "@" + sha
```
Jefferson-Faseler, line 63: It's picky, but `fmt.Sprintf`?

cmd/root.go
```diff
@@ -29,20 +29,17 @@ func init() {
...
-		}
 	} else {
 		logger, err = zap.NewProduction()
-		if err != nil {
-			return
-		}
 	}
 
-	return
+	return logger, err
```
Jefferson-Faseler, line 42: It's the little things. 💯

cmd/start.go
```diff
@@ -42,100 +41,83 @@ func bindShutdown(logger *zap.Logger, cancel func()) {
...
-		Unprocessed:       unprocessedEvents,
-		Processed:         processedEvents,
-		ChErr:             failedEvents,
-		ChErrReports:      errReports,
+	if viper.GetBool("reviewapps-enabled") {
+		go startReviewAppsServer(logger, creds)
 	}
-	go eventProcessor.Start()
 
-	go func() {
```
Jefferson-Faseler, line 117: 👋

pkg/containers/repository.go
```diff
@@ -38,7 +38,7 @@ func (a apiError) Error() string {
 type manifest struct {
 	Digest string
 	Layers []layer    `json:"layers"`
-	Errors []apiError `json:"errors"`
+	Errors []apiError `json:"report"`
```
Jefferson-Faseler, line 41: How long was this incorrect? SMH

Jefferson-Faseler (commented):
How should I test this?

pkg/containers/repository.go
jkminneti: oh god good catch, that's not incorrect

cmd/deploy.go
jkminneti:
```
fmt.Sprintf("%s/%s@%s", app.RepoHost, app.RepoPath, sha)
```
is that really clearer?
I'm not opposed cus i think that's closer to the convention but that's just not clearer to me

Jefferson-Faseler: I mean. At this point we are all at least 10% computers. But yeah, you got a point. I'll let you follow your heart.

PR #158, opened 2020-11-06
POPS-698 Pause and Unpause Releases

Hbbb:
## What does this PR Do
* Adds `pause [app]` and `resume [app]` commands 
* Discards deployments if an app is paused

The app pauses are stored in a configmap that looks like this:
```yaml
# tuber-app-pauses
data:
  account-service: true
```

Pausing and unpausing apps adds/removes apps from the configmap.

### DIFF
--- a/pkg/core/deployments.go
+++ b/pkg/core/deployments.go
@@ -2,9 +2,11 @@ package core
 
 import "tuber/pkg/k8s"
 
+const tuberAppPauses = "tuber-app-pauses"
+
 // ReleasesPaused checks the tuber-app-pauses configmap for the presence of the given app
 func ReleasesPaused(appName string) (bool, error) {
-	config, err := k8s.GetConfigResource("tuber-app-pauses", "tuber", "ConfigMap")
+	config, err := k8s.GetConfigResource(tuberAppPauses, "tuber", "ConfigMap")
 	if err != nil {
 		return false, err
 	}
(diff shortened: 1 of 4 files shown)


pkg/events/processor.go
```diff
@@ -57,6 +57,17 @@ func (p Processor) ProcessMessage(digest string, tag string) {
 	for _, app := range apps {
 		if app.ImageTag == event.tag {
 			matchFound = true
+
+			paused, err := core.ReleasesPaused(app.Name)
```
jkminneti, line 61: that'll shadow the outer `err`

pkg/core/deployments.go
```diff
@@ -0,0 +1,43 @@
...
+	exists, err := k8s.Exists("configmap", "tuber-app-pauses", "tuber")
+	if err != nil {
+		return err
+	}
+
+	if !exists {
+		return nil
+	}
+
+	return k8s.RemoveConfigMapEntry("tuber-app-pauses", "tuber", appName)
```
jkminneti, line 42: `"tuber-app-pauses"` should maybe be a constant

pkg/events/processor.go
```diff
@@ -57,6 +57,17 @@ func (p Processor) ProcessMessage(digest string, tag string) {
...
+			if err != nil {
+				event.logger.Error("failed to check for paused state", zap.Error(err))
+			}
+
+			if paused {
+				event.logger.Warn("app deployments paused; skipping", zap.Bool("paused", true))
```
jkminneti, line 67: `zap.Bool("paused", true)` doesn't seem all that useful given the message, and the context we ARE missing is the app name

pkg/events/processor.go
Hbbb: Outer `err`? The previous definition is on line 48, and if it's not `nil` the function returns.

PR #159, opened 2020-11-06
POPS-665 Block Simultaneous Releases

Hbbb:
[POPS-665](https://freshly.atlassian.net/browse/POPS-665)

The current strategy is to use a `map[string]*sync.Mutex` which gives us a mutex per app. Each deploy event will lock the app-specific mutex, preventing other deployments of the same app until the goroutine that obtained the lock releases it.

Here are logs samples from the testing I ran locally:

### Blocking Simultaneous Releases:

```
{release starting", "ts":1605640778.19966,  "digest":"gcr.io/freshly-docker/potatoes@sha256:AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA","name":"potatoes-time-heist","branch":"time-heist","imageTag":"gcr.io/freshly-docker/potatoes:time-heist","action":"release"}
{release complete", "ts":1605640807.536586, "digest":"gcr.io/freshly-docker/potatoes@sha256:AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA","name":"potatoes-time-heist","branch":"time-heist","imageTag":"gcr.io/freshly-docker/potatoes:time-heist","action":"release","duration":26.781038628}
{release starting", "ts":1605640807.536752, "digest":"gcr.io/freshly-docker/potatoes@sha256:BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB","name":"potatoes-time-heist","branch":"time-heist","imageTag":"gcr.io/freshly-docker/potatoes:time-heist","action":"release"}
{release complete", "ts":1605640834.523639, "digest":"gcr.io/freshly-docker/potatoes@sha256:BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB","name":"potatoes-time-heist","branch":"time-heist","imageTag":"gcr.io/freshly-docker/potatoes:time-heist","action":"release","duration":24.792363043}
{release starting", "ts":1605640834.523774, "digest":"gcr.io/freshly-docker/potatoes@sha256:CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC","name":"potatoes-time-heist","branch":"time-heist","imageTag":"gcr.io/freshly-docker/potatoes:time-heist","action":"release"}
{release complete", "ts":1605640863.182157, "digest":"gcr.io/freshly-docker/potatoes@sha256:CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC","name":"potatoes-time-heist","branch":"time-heist","imageTag":"gcr.io/freshly-docker/potatoes:time-heist","action":"release","duration":26.52911669}
{release starting", "ts":1605640863.182295, "digest":"gcr.io/freshly-docker/potatoes@sha256:DDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDD","name":"potatoes-time-heist","branch":"time-heist","imageTag":"gcr.io/freshly-docker/potatoes:time-heist","action":"release"}
{release complete", "ts":1605640891.799352, "digest":"gcr.io/freshly-docker/potatoes@sha256:DDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDD","name":"potatoes-time-heist","branch":"time-heist","imageTag":"gcr.io/freshly-docker/potatoes:time-heist","action":"release","duration":26.688430478}
```

### Not Blocking Simultaneous Releases:
```
{release starting","digest":"gcr.io/freshly-docker/potatoes@sha256:AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA","name":"potatoes-time-heist","branch":"time-heist","imageTag":"gcr.io/freshly-docker/potatoes:time-heist","action":"release"}
{release starting","digest":"gcr.io/freshly-docker/potatoes@sha256:BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB","name":"potatoes-time-heist","branch":"time-heist","imageTag":"gcr.io/freshly-docker/potatoes:time-heist","action":"release"}
{release starting","digest":"gcr.io/freshly-docker/potatoes@sha256:CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC","name":"potatoes-time-heist","branch":"time-heist","imageTag":"gcr.io/freshly-docker/potatoes:time-heist","action":"release"}
{release starting","digest":"gcr.io/freshly-docker/potatoes@sha256:DDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDD","name":"potatoes-time-heist","branch":"time-heist","imageTag":"gcr.io/freshly-docker/potatoes:time-heist","action":"release"}
{release starting","digest":"gcr.io/freshly-docker/potatoes@sha256:EEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEE","name":"potatoes-time-heist","branch":"time-heist","imageTag":"gcr.io/freshly-docker/potatoes:time-heist","action":"release"}
{release complete","digest":"gcr.io/freshly-docker/potatoes@sha256:BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB","name":"potatoes-time-heist","branch":"time-heist","imageTag":"gcr.io/freshly-docker/potatoes:time-heist","action":"release","duration":31.143002553}
{release complete","digest":"gcr.io/freshly-docker/potatoes@sha256:DDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDD","name":"potatoes-time-heist","branch":"time-heist","imageTag":"gcr.io/freshly-docker/potatoes:time-heist","action":"release","duration":30.127527718}
{release complete","digest":"gcr.io/freshly-docker/potatoes@sha256:CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC","name":"potatoes-time-heist","branch":"time-heist","imageTag":"gcr.io/freshly-docker/potatoes:time-heist","action":"release","duration":30.15295781}
{release complete","digest":"gcr.io/freshly-docker/potatoes@sha256:AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA","name":"potatoes-time-heist","branch":"time-heist","imageTag":"gcr.io/freshly-docker/potatoes:time-heist","action":"release","duration":31.404309026}
{release complete","digest":"gcr.io/freshly-docker/potatoes@sha256:EEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEE","name":"potatoes-time-heist","branch":"time-heist","imageTag":"gcr.io/freshly-docker/potatoes:time-heist","action":"release","duration":30.352348723}
```


pkg/events/processor.go
```diff
@@ -53,13 +58,27 @@ func (p Processor) ProcessMessage(digest string, tag string) {
...
+		p.locks[event.tag] = &sync.Mutex{}
+	}
+	p.ll.Unlock()
+
 	matchFound := false
+
+	// We have to lock the mutex here because we have to unlock the mutex at the end of the for loop,
+	// after all of the apps matching the event.tag have deployed.
+	mu := p.locks[event.tag]
+	mu.Lock()
```
jkminneti, line 72: so, same point as I was saying last week - there's not really any value in doing this for every incoming tag. 

still thinking we should filter first, then if matches are found do the lock, then loop through the matches and run releases

jkminneti (changes requested):

pkg/events/processor.go
```diff
@@ -17,6 +18,8 @@ type Processor struct {
 	creds             []byte
 	clusterData       *core.ClusterData
 	reviewAppsEnabled bool
+	locks             map[string]*sync.Mutex
```
jkminneti, line 21: probably should be `*map[string]*sync.Mutex`. no reason not to that i can tell, and leaving it as a value means the processor _can never be copied_. which is sorta fine as there's no reason to copy it, but again no reason not to allow it.

jkminneti (commented):
also sorta thinking this could be branched off of release pausing (or one big branch) just to avoid merge conflicts but 🤷

pkg/events/processor.go
Hbbb: Ok, question about this. In a situation like with potatoes, where a single `event.tag` will match several apps, if we obtain the lock in the beginning of the match block, the subsequent deploys will hang. This is the only workable way I can think of:

```go
for _, app := range apps {
	if app.ImageTag == event.tag {
		matchFound = true

		mu := p.locks[event.tag]
		mu.Lock()

		p.startRelease(event, &app)
		mu.Unlock()
	}
}
```

Is this what you're thinking?

jkminneti: im saying filter and on match, append to an array
then if len > 0, lock, loop over em and run releases

pkg/events/processor.go
```diff
@@ -53,13 +60,30 @@ func (p Processor) ProcessMessage(digest string, tag string) {
...
 	for _, app := range apps {
 		if app.ImageTag == event.tag {
 			matchFound = true
-			p.startRelease(event, &app)
+			deployments = append(deployments, &app)
 		}
 	}
+
+	if len(deployments) > 0 {
+		mu := (*p.locks)[event.tag]
```
jkminneti, line 79: `(*p.locks)` doesn't have to be in parenthesis does it? or is there some magic there?

pkg/events/processor.go
```diff
@@ -53,13 +60,30 @@ func (p Processor) ProcessMessage(digest string, tag string) {
 	}
 	event.logger.Debug("filtering event against current tuber apps", zap.Any("apps", apps))
 
+	p.ll.Lock()
+	if _, ok := (*p.locks)[event.tag]; !ok {
+		(*p.locks)[event.tag] = &sync.Mutex{}
+	}
+	p.ll.Unlock()
```
jkminneti, line 67: seems like this stuff can and should move under the `if len(deployments) > 0 {` block

pkg/events/processor.go
```diff
@@ -53,13 +60,30 @@ func (p Processor) ProcessMessage(digest string, tag string) {
...
 		if app.ImageTag == event.tag {
 			matchFound = true
-			p.startRelease(event, &app)
+			deployments = append(deployments, &app)
 		}
 	}
+
+	if len(deployments) > 0 {
+		mu := (*p.locks)[event.tag]
+		mu.Lock()
```
jkminneti, line 80: actually oof. 
```
3 deploys come in
first one locks here
2nd and 3rd get blocked
first one unlocks
2nd and 3rd proceed at the same time
```
is that incorrect?
if not, that's where `sync.Cond` comes in i think

Hbbb: I think you're wrong. Here's what I've got:

```go
func monitor(wg *sync.WaitGroup, ch chan string) {
	wg.Wait()
	close(ch)
}

func main() {
	m := &sync.Mutex{}
	wg := &sync.WaitGroup{}
	messages := make(chan string)

	for i := 0; i < 4; i++ {
		wg.Add(1)

		go func(ch chan string, w *sync.WaitGroup, idx int) {
			ch <- fmt.Sprintf("worker %d - LOCK", idx)
			m.Lock()

			ch <- fmt.Sprintf("worker %d - **WORK**", idx)
			time.Sleep(1 * time.Second)

			ch <- fmt.Sprintf("worker %d - UNLOCK", idx)
			m.Unlock()
			w.Done()
		}(messages, wg, i)
	}

	go monitor(wg, messages)

	for msg := range messages {
		fmt.Println(msg)
	}
}
```

Here's a super dumbed down example of what we've got. Several goroutines start, request a lock on the mutex, "do work", then unlock the mutex. 

Here are the log outputs:
```
worker 0 - LOCK
worker 0 - **WORK**
worker 1 - LOCK
worker 2 - LOCK
worker 3 - LOCK
worker 0 - UNLOCK
worker 2 - **WORK**
worker 2 - UNLOCK
worker 1 - **WORK**
worker 1 - UNLOCK
worker 3 - **WORK**
worker 3 - UNLOCK
```

Is this a fair test?

jkminneti: i think that proves it... no?
worker 2 does work before worker 1 there
which makes me think they're _all_ unlocked at the same time to continue

Hbbb: Yea, you're right. They still execute one at a time, but when the first goroutine unlocks, any of the paused goroutines can get the next lock.

Hbbb:
Ok @jkminneti. Finally updated and got it working. I updated the logs in the PR description to reflect the most recent test runs. They're kind of messy but they illustrate the point. The deploys execute in order, one at a time.

jaysonesmith:
@Hbbb I'm sorry about being quiet on this. I'm not sure how to test it locally to ensure it's working appropriately.

Hbbb:
@jaysonesmith Do you want to pair and test this tomorrow? Also, any feedback on the code is much appreciated 😄

pkg/events/processor.go
```diff
@@ -54,23 +59,41 @@ func (p Processor) ProcessMessage(digest string, tag string) {
...
+			(*p.locks)[event.tag] = sync.NewCond(&mutex)
+		}
+
+		cond := (*p.locks)[event.tag]
+		cond.L.Lock()
+		for _, app := range deployments {
 			p.startRelease(event, &app)
 		}
+		cond.L.Unlock()
+		cond.Signal()
```
jkminneti, line 94: i think the paused check should be within this loop - if a release ends up getting queued and the app is paused before the deploy starts, it'll continue deploying

another thought is - an app on the same tag as another app could block the other app's deployments (because the cond is off the tag, not the app name). not sure that.. matters though?

jkminneti: like functionally i don't think this matters - but in principle we're blocking simultaneous _app_ deploys, not tag deploys.

if there's an actual functional difference though, we'd just need to (base it off the app name) and send out goroutines that actually call the `cond.L.Lock`

jkminneti: and the obvious functional difference is that the other app would be blocked lol. but my thought is that timeouts should take care of that for the most part

pkg/events/processor.go
```diff
@@ -54,23 +59,41 @@ func (p Processor) ProcessMessage(digest string, tag string) {
...
 
+			deployments = append(deployments, app)
+		}
+	}
+
+	if len(deployments) > 0 {
+		if _, ok := (*p.locks)[event.tag]; !ok {
+			var mutex sync.Mutex
+			(*p.locks)[event.tag] = sync.NewCond(&mutex)
+		}
```
jkminneti, line 86: does feel like
```go
if (*p.locks)[event.tag] != nil {
```
would be more readable but 🤷

quinn-freshly (approved):

pkg/events/processor.go
quinn-freshly: i agree with this. it seems more appropriate to do it off app name. this probably would never come up, but it would at least make the intent of the code more clear

Hbbb: This is a great point that I hadn't considered. It definitely needs to be based off appname.

Hbbb: Ok it just occurred to me, can't we just do this:
```go
if len(deployments) > 0 {
	for _, app := range deployments {
		if _, ok := (*p.locks)[app.Name]; !ok {
			var mutex sync.Mutex
			(*p.locks)[app.Name] = sync.NewCond(&mutex)
		}

		cond := (*p.locks)[app.Name]
		cond.L.Lock()

		paused, err := core.ReleasesPaused(app.Name)
		if err != nil {
			event.logger.Error("failed to check for paused state; deploying", zap.Error(err))
		}

		if paused {
			event.logger.Warn("deployments are paused for this app; skipping", zap.String("appName", app.Name))
			continue
		}
		p.startRelease(event, &app)
		cond.L.Unlock()
		cond.Signal()
	}
}
```

jkminneti: so that would block deployments on a given app, but while waiting for deploys of a given tag. 

- message 1 received
- app 1 starts deploying
- app 2 sits there waiting
- message 2 received
- app 2 hasn't locked yet, deploys immediately

that's why i was saying the loop should fire off goroutines that call the lock, and run all releases for a given message at the same time

jkminneti: should note though - message 2 would likely try to deploy app 1 first, just like message 1, and get locked until app 2's first deploy all the same.

so unless you changed your app's tag in between messages, that would actually kinda work. 

but, seems like it's hanging on a loophole

pkg/events/processor.go
```diff
@@ -54,23 +59,43 @@ func (p Processor) ProcessMessage(digest string, tag string) {
...
-			if paused {
-				event.logger.Warn("app deployments paused; skipping", zap.String("appName", app.Name))
-				continue
-			}
+			deployments = append(deployments, app)
+		}
+	}
 
-			p.startRelease(event, &app)
+	if len(deployments) > 0 {
```
jkminneti, line 72: realized something - now that the entire block to handle a match is off thread, I don't think we need to loop through to find the matches ahead of time

jkminneti (approved):

PR #160, opened 2020-11-10
Add Local Testing Guide

Hbbb:


doc/testing/testing_locally.md
```diff
@@ -0,0 +1,67 @@
...
+git checkout -b super-test
+```
+
+Install test version of potatoes app, using your branch as the final argument:
+```bash
+tuber apps install potatoes-super-test us.gcr.io/freshly-docker/potatoes super-test
+```
+
+## Trigger Pubsub Events
+There are several ways to trigger events that will cause tuber to re-build your app. The potatoes cloud build trigger is configured to build on pushes to any branch, so it's ideal for testing.
```
jkminneti, line 35: this is OSS - no potatoes, no links to our gcp resources, etc.

jkminneti (changes requested):

doc/testing/testing_locally.md
jkminneti: if we want functional prebuilt stuff in the docs, we need a dedicated private repo for tuber stuff (which may be ideal anyway for tuber configs)

Hbbb: Maybe this makes more sense as a handbook entry? I my focus is not on a generic testing guide, it's more about enabling developers at Freshly to be able to work with tuber.

Hbbb:
I've moved this guide to the handbook.

PR #161, opened 2020-11-24
pull tuber files from image tag, not latest

jkminneti:


Hbbb (approved):
This looks good, but can you remind me what was the result of this bug again?

PR #171, opened 2021-02-08
remove test code and pull business logic out of controllers

jkminneti:


pkg/reviewapps/reviewapps.go
```diff
@@ -28,6 +32,104 @@ func NewReviewAppSetup(sourceApp string, reviewApp string) error {
...
+
+		if cleanupConfigErr != nil {
+			logger.Error("error removing config entry for app", zap.Error(cleanupConfigErr))
+			return cleanupConfigErr
+		}
+
+		if triggerCleanupErr != nil {
+			logger.Error("error removing trigger", zap.Error(triggerCleanupErr))
+			return triggerCleanupErr
+		}
```
jaysonesmith, line 121: Are these handled here so that we can make sure we clean up all state first? Is it possible and would we care about instances where all of or two out of the three errors aren't nil?

Currently it looks like we'd only report whatever the first error was and lose out on info of any others?

jaysonesmith: Also does order of operations for clean up matter? I just noticed that in `DeleteReviewApp` the steps are in a different orded.

jaysonesmith (commented):
All for making handlers as minimal as possible and passing things off to other functions. 🎉

pkg/reviewapps/server.go
```diff
@@ -123,39 +34,23 @@ func (s *Server) CreateReviewApp(ctx context.Context, in *proto.CreateReviewAppR
...
-	if err != nil {
-		logger.Error("error deleting tuber app", zap.Error(err))
-		return nil, err
-	}
+	err := DeleteReviewApp(ctx, reviewAppName, s.Credentials, s.ProjectName)
 
-	err = core.RemoveReviewAppConfig(reviewAppName)
 	if err != nil {
-		logger.Error("error deleting tuber review map app config entry item row", zap.Error(err))
+		logger.Error("error deleting review app " + reviewAppName + ": " + err.Error())
```
jaysonesmith, line 46: Linter complains here about the use of `reviewAppName` because in this setting it's a function call and not a variable, like down in `DeleteReviewApp`: `Sprintf format %s arg reviewAppName is a func value, not called`

pkg/reviewapps/permissions.go
```diff
@@ -15,9 +15,7 @@ func canCreate(logger *zap.Logger, appName, token string) (bool, error) {
 
 	canDeploy := k8s.CanDeploy(appName, token)
 
-	return (appName != "tuber" &&
-		canDeploy &&
-		exists), nil
+	return appName != "tuber" && canDeploy && exists, nil
```
jaysonesmith, line 18: How much do we care about magic strings? In just the `./pkg/reviewapps` package it looks like we use the bare string `"tuber"` in arguments four times and as a conditional the one time here. In a couple of the argument cases, we're already using string constants instead of the magic string.

Not 100% related to magic strings: I _think_ from reading the code that this return value is to at least partially prevent a review app also called `"tuber"` being deployed and stomping on actual tuber?

What are your thoughts about pulling the `appName` check out and running it at the top of the function as a guard clause?

```go
	if appName == "tuber" {
		return false, nil
	}
```

If my above understanding is correct, then returning fast seems like it'd make a lot of sense. It'd prevent having to bother making three requests to kubectl that we wouldn't care about if the app name is tuber.

cmd/start.go
```diff
@@ -105,10 +106,25 @@ func start(cmd *cobra.Command, args []string) {
 func startReviewAppsServer(logger *zap.Logger, creds []byte) {
 	logger = logger.With(zap.String("action", "grpc"))
 
+	defaultHost := viper.GetString("cluster-default-host")
+	if defaultHost == "" {
+		err := fmt.Errorf("no cluster default host configured")
+		logger.Error("grpc server failed to start: no cluster default host configured")
+		report.Error(err, report.Scope{"during": "grpc server startup"})
+		panic(err)
+	}
```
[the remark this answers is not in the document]
jkminneti: you know what - this behavior is salvageable without one configured. ill edit this out.

for context, the `clusterData` is designed to use local env and fall back to _the running tuber's env_. you use it regularly without realizing it.

all part of the project of "if you have access to the tuber namespace, what if actually zero env vars were required for the local tuber binary"

but yeah for this there IS no fallback. all we have is the running tuber's env, and there's generally and fundamentally no fallback value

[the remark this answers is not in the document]
jkminneti: also you'd be kinda insane NOT to use cluster default host but whatever we can make it less explosive

cmd/start.go
```diff
@@ -105,10 +106,25 @@ func start(cmd *cobra.Command, args []string) {
...
+
+	projectName := viper.GetString("review-apps-triggers-project-name")
```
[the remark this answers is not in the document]
jkminneti: `review apps` is the feature though? and the project name indicates which project for _this specific thing_ and results in resources getting created and destroyed there. 

in all other ways tuber does not care where you do your builds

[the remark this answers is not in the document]
jkminneti: point is, this is for trigger creation now, and should ALWAYS be for trigger creation

pkg/reviewapps/reviewapps.go
jkminneti: zero idea - this just copy pastes it out of the controller functions

pkg/reviewapps/server.go
jkminneti: ifix

pkg/reviewapps/permissions.go
jkminneti: yup fixed. just wanted to get rid of those parentheses but now youve inspired me to actually refactor it lol.

i care less about the const vs hardcoded stuff though, we can trim that down later.

pkg/reviewapps/client.go
```diff
@@ -1,20 +1,20 @@
...
 // NewClient returns a GRPC client
 func NewClient(url string) (proto.TuberClient, *grpc.ClientConn, error) {
-	hostname := ":3000" //fmt.Sprintf(":9000", url)
-
-	var conn *grpc.ClientConn
-	// creds := credentials.NewTLS(&tls.Config{})
-
-	conn, err := grpc.Dial(hostname, grpc.WithInsecure())
+	config := &tls.Config{
+		InsecureSkipVerify: true,
```
quinn-freshly, line 15: why is this necessary

pkg/reviewapps/permissions.go
```diff
@@ -8,16 +8,16 @@ import (
...
+	if err != nil || !exists {
 		return false, err
 	}
 
-	canDeploy := k8s.CanDeploy(appName, token)
-
-	return (appName != "tuber" &&
-		canDeploy &&
-		exists), nil
+	return k8s.CanDeploy(appName, token), nil
```
quinn-freshly, line 20: nice

pkg/reviewapps/reviewapps.go
```diff
@@ -28,6 +33,117 @@ func NewReviewAppSetup(sourceApp string, reviewApp string) error {
...
+			return "", teardownErr
+		}
+
+		return "", err
+	}
+
+	logger.Info("creating app entry for review app")
+
+	err = core.AddReviewAppConfig(reviewAppName, sourceApp.Repo, branch)
+	if err != nil {
```
quinn-freshly, line 89: maybe add  the "tearing down" leg message here too?

pkg/reviewapps/permissions.go
jaysonesmith: 🎉

pkg/reviewapps/client.go
jkminneti: fair, ill try without - artifact from getting it working

jaysonesmith (approved):
Assuming the removal of `InsecureSkipVerify` works, lgtm ✅

PR #172, opened 2021-02-22
client side can-i

jkminneti:
client-side validations aren't just generally nice, they're also required in this case - if a user's token is stale, nothing in the existing review apps creation process refreshes it, so we're sending the expired token for server validation. this does the same checks, and in doing so, refreshes the user's token

### DIFF
--- a/cmd/reviewApps.go
+++ b/cmd/reviewApps.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"tuber/pkg/core"
 	"tuber/pkg/k8s"
 	"tuber/pkg/proto"
 	"tuber/pkg/reviewapps"
@@ -50,6 +51,19 @@ func create(cmd *cobra.Command, args []string) error {
 		return err
 	}
 	defer conn.Close()
+
+	_, err = core.FindApp(sourceAppName)
+	if err != nil {
+		return fmt.Errorf("source app not found")
+	}
+
+	canDeploy, err := k8s.CanDeploy(sourceAppName)
+	if err != nil {
+		return err
+	}
+	if !canDeploy {
+		return fmt.Errorf("not permitted to create a review app from %s", sourceAppName)
+	}
 
 	config, err := k8s.GetConfig()
 	if err != nil {
--- a/cmd/reviewApps.go
+++ b/cmd/reviewApps.go
@@ -19,16 +19,16 @@ var reviewAppsCmd = &cobra.Command{
 var reviewAppsCreateCmd = &cobra.Command{
 	SilenceUsage: true,
 	Use:          "create [source app name] [branch name]",
-	Short:        "",
+	Short:        "Create a temporary application deployed alongside the source application for a given branch, copying its rolebindings and env",
 	Args:         cobra.ExactArgs(2),
 	RunE:         create,
 }
 
 var reviewAppsDeleteCmd = &cobra.Command{
 	SilenceUsage: true,
-	Use:          "delete [review app name]",
-	Short:        "",
-	Args:         cobra.ExactArgs(1),
+	Use:          "delete [source app name] [branch]",
+	Short:        "Delete a review app",
+	Args:         cobra.ExactArgs(2),
 	RunE:         delete,
 }
 
@@ -43,7 +43,7 @@ func create(cmd *cobra.Command, args []string) error {
 
 	clusterConf := tuberConf.CurrentClusterConfig()
 	if clusterConf.URL == "" {
-		return fmt.Errorf("no cluster config found. run `tuber config`")
+		return fmt.Errorf("no tuber url found for current cluster. run `tuber config`")
 	}
 
 	client, conn, err := reviewapps.NewClient(clusterConf.URL)
@@ -92,7 +92,9 @@ func create(cmd *cobra.Command, args []string) error {
 }
 
 func delete(cmd *cobra.Command, args []string) error {
-	reviewAppName := args[0]
+	sourceAppName := args[0]
+	branch := args[1]
+	reviewAppName := reviewapps.ReviewAppName(sourceAppName, branch)
 
 	tuberConf, err := getTuberConfig()
 	if err != nil {
@@ -101,7 +103,7 @@ func delete(cmd *cobra.Command, args []string) error {
 
 	clusterConf := tuberConf.CurrentClusterConfig()
 	if clusterConf.URL == "" {
-		return fmt.Errorf("no cluster config found. run `tuber config`")
+		return fmt.Errorf("no tuber url found for current cluster. run `tuber config`")
 	}
 
 	client, conn, err := reviewapps.NewClient(clusterConf.URL)
@@ -109,6 +111,11 @@ func delete(cmd *cobra.Command, args []string) error {
 		return err
 	}
 	defer conn.Close()
+
+	_, err = core.FindReviewApp(reviewAppName)
+	if err != nil {
+		return fmt.Errorf("review app not found")
+	}
 
 	config, err := k8s.GetConfig()
 	if err != nil {
--- a/pkg/k8s/kubectl.go
+++ b/pkg/k8s/kubectl.go
@@ -2,7 +2,6 @@ package k8s
 
 import (
 	"encoding/json"
-	"fmt"
 	"os"
 	"os/exec"
 	"strings"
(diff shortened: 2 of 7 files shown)


pkg/reviewapps/permissions.go
```diff
@@ -1,14 +1,15 @@
...
+	"fmt"
 	"tuber/pkg/core"
 	"tuber/pkg/k8s"
 
 	"go.uber.org/zap"
 )
 
 func canCreate(logger *zap.Logger, appName, token string) (bool, error) {
-	if appName == "tuber" {
+	if appName == "tuber" || token == "" {
```
jaysonesmith, line 12: unrelated to this change, but are we passing in a logger as a future plan sort of thing? It's not currently used in this function.

pkg/reviewapps/permissions.go
```diff
@@ -17,7 +18,12 @@ func canCreate(logger *zap.Logger, appName, token string) (bool, error) {
 		return false, err
 	}
 
-	return k8s.CanDeploy(appName, token), nil
+	canDeploy, err := k8s.CanDeploy(appName, fmt.Sprintf("--token=%s", token))
+	if err != nil {
+		return false, err
+	}
+
+	return canDeploy, nil
```
jaysonesmith, line 26:
```suggestion
	return k8s.CanDeploy(appName, fmt.Sprintf("--token=%s", token))
```

`k8s.CanDeploy` return values satisfy `canCreate`'s and if `canCreate` errors, it already returns `false, err` so you're just overwriting `canDeploy` on line 23.

pkg/k8s/kubectl.go
```diff
@@ -205,19 +204,14 @@ func UseCluster(cluster string) error {
...
+func CanDeploy(namespace string, args ...string) (bool, error) {
+	canDeploy := []string{"auth", "can-i", "create", "deployments", "-n", namespace}
+	out, err := kubectl(append(canDeploy, args...)...)
 	if err != nil {
-		return false
+		return false, err
 	}
 
-	return strings.Trim(string(out), "\r\n") == "yes"
+	return strings.Trim(string(out), "\r\n") == "yes", nil
```
jaysonesmith, line 214: You can get rid of the err check here and just end with.

`return strings.Trim(string(out), "\r\n") == "yes", err`

If out is a nil slice of bytes, it'll come out as an empty string through the conversion.

pkg/reviewapps/permissions.go
jaysonesmith: I still really love that this early guard is in place. It feels so nice getting out of the function early and not calling all the other stuff.

cmd/reviewApps.go
```diff
@@ -51,6 +52,19 @@ func create(cmd *cobra.Command, args []string) error {
 	}
 	defer conn.Close()
 
+	_, err = core.FindApp(sourceAppName)
+	if err != nil {
+		return fmt.Errorf("source app not found")
+	}
+
+	canDeploy, err := k8s.CanDeploy(sourceAppName)
```
jaysonesmith, line 60: I assume this is what you're using to do the client side check, but I'm not sure how this would end up refreshing someone's token. Could you explain the process, please?

cmd/reviewApps.go
```diff
@@ -51,6 +52,19 @@ func create(cmd *cobra.Command, args []string) error {
 	}
 	defer conn.Close()
 
+	_, err = core.FindApp(sourceAppName)
```
jaysonesmith, line 55: _this comment may all not matter if there are folks who are/could be calling tuber outside of the cli by making direct calls via grpc or something_

If we're moving this check client side (which I totally approve of as it prevents unnecessary calls to the server), do we also need to perform this same check in `core.CreateReviewApp`? It's potentially not a huge deal, but for this particular example, it'd cut down on one call to k8s since it's already happening client side.

On that same thread, should we be moving the check for [a preexisting review app](https://github.com/Freshly/tuber/blob/client-side-can-i/pkg/reviewapps/reviewapps.go#L39-L46) client side as well?

jaysonesmith: _continuing a similar thread as above, but with an additional angle. it also may not matter at all_

`core.CreateReviewApp` also calls `canCreate` which under the hood makes three calls to k8s:

* core.SourceAndReviewApps which calls the following
    * core.TuberSourceApps - Called again later on line 66
    * core.TuberReviewApps - Already called in function execution on line 39
* k8s.CanDeploy - Called client side

Of those calls TuberReviewApps will have already been called on line 39, `TuberSourceApps` will be called again on line 66, and then there's the question of CanDeploy being duplicative of a _similar_ call client side.

From what I can tell, there's no CRUD that happens in between the two TuberXXXApps calls that would cause data to go stale, so we could call them once at the top of the function and use their data throughout the rest of the function. This could actually replace `canDeploy` as a function and simplify other parts of `CreateReviewApp`.

I had a bunch more text written up, but this is probably better.

https://play.golang.org/p/JcU9Q9WoohO

jaysonesmith (approved):
Aside from one comment I've left asking for more info, the rest are responsibility/style/efficiency things mainly.

I imagine it's something on my end that isn't configured correctly, but I also get the following locally. Is this known/expected?

```
╭─16:18 jaysonsmith@PHX2-MACLT-2004 ~/dev/code/freshly/tuber ‹client-side-can-i*›
╰─> go run ./main.go review-apps create potatoes staging
Error: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp: address https://tuber.internal.freshlyservices.net/:443: too many colons in address"
```

pkg/k8s/kubectl.go
jkminneti: still a behavior change - who's to say it won't error AND return "yes".

ok common sense is who's to say, but still - i think errors should short circuit logic and return

cmd/reviewApps.go
jkminneti: the server must have zero trust in the client, so duplicative checks are kinda the point here? i may be missing something.

cmd/reviewApps.go
jkminneti: kubectl locally holds tokens and config for each configured cluster. the token, if expired, is re-authed automatically by kubectl when making a remote request. so as long as we're doing at least 1 connection to the cluster before preparing the grpc request, we're good

try `kubectl config view` if you're curious what that looks like

jkminneti:
> Aside from one comment I've left asking for more info, the rest are responsibility/style/efficiency things mainly.
> 
> I imagine it's something on my end that isn't configured correctly, but I also get the following locally. Is this known/expected?
> 
> ```
> ╭─16:18 jaysonsmith@PHX2-MACLT-2004 ~/dev/code/freshly/tuber ‹client-side-can-i*›
> ╰─> go run ./main.go review-apps create potatoes staging
> Error: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp: address https://tuber.internal.freshlyservices.net/:443: too many colons in address"
> ```

2 issues there - you're on internal so no potatoes, and your `tuber config` is outdated. i can post the updated one in slack

pkg/k8s/kubectl.go
jaysonesmith: It certainly could return yes & an error but it's the job of the caller to check the error before acting on the bool value. :) I think the only real reason I'd want to leave this is to avoid the conversion if that was something we were worried about as far as performance goes.

Ultimately, the byte slice to string conversion and trim is extremely fast: 

`BenchmarkConvertRes-12    	14126377	        78.2 ns/op	      48 B/op	       2 allocs/op`

cmd/reviewApps.go
jaysonesmith: Nah, that's totally valid inre validation actuall. I still think that though, that restructuring the order of the calls to minimize extra/repeated calls would be a good idea.

cmd/reviewApps.go
jaysonesmith: Ah I get the behavior now. 👍

jaysonesmith (approved):
My $0.02 are in the comments but ultimately it all appears like it'll work regardless. ✅

jaysonesmith:
> 2 issues there - you're on internal so no potatoes, and your `tuber config` is outdated. i can post the updated one in slack

I've got the update tuber config ty!

I tried some more local runs with potatoes and ran into some unexpected responses. Unsure if it's a PEBKAC issue with my config or understanding, but I'm gonna chuck some stuff in a gist and send to you.

PR #174, opened 2021-03-09
admin dash

jkminneti:
let's just keep in mind this is an MVP eh

edit: nah screw this, see #178


.tuber/virtualservice.yaml
```diff
@@ -4,9 +4,9 @@ metadata:
   name: tuber
 spec:
   hosts:
-  - "tuber.{{.clusterDefaultHost}}"
+    - "tuber.{{.clusterDefaultHost}}"
   gateways:
-  - "{{.clusterDefaultGateway}}"
+    - "{{.clusterDefaultGateway}}"
```
quinn-freshly, line 9: can we keep this indentation standard? its same-line in most other plsaces

quinn-freshly (changes requested):

.tuber/virtualservice.yaml
quinn-freshly: this is one of my least favorite aspects of yaml

pkg/adminserver/dashboard.go
```diff
@@ -0,0 +1,69 @@
...
+	return dashboard{
+		routePath: "/",
+		raw: `<head><title>Tuber Admin</title></head>
+<h1>potate</h1>
+{{if .Error}}
+	<p>{{.Error}}</p>
+{{else}}
+	<table>{{range .SourceApps}}
+		<tr>
+			<td><a href="{{$.AppPath}}/{{.Name}}">{{.Name}}</a></td>
```
quinn-freshly, line 20: how would the codebase change if $.AppPath became a hard-coded value? I think it would reduce complexity a bit without really losing much

quinn-freshly (changes requested):

pkg/adminserver/dashboard.go
jkminneti: hardcoded in the html or?

quinn: yeah, like `href="/apps/{{`

quinn: can you mark this as resolved? its not blocking

jkminneti: sure. as for how it changes the codebase - if we want that to change to `source_apps/` or something, we'd then have to change it in >1 places

cmd/adminserver.go
```diff
@@ -0,0 +1,46 @@
...
+func startAdminServer(cmd *cobra.Command, args []string) {
+	var ctx, cancel = context.WithCancel(context.Background())
+	defer cancel()
+
+	logger, err := createLogger()
+	defer logger.Sync()
+
+	if err != nil {
+		panic(err)
+	}
```
jaysonesmith, line 27: In this particular case it doesn't matter, but I think it's probably best practice to move defers after error checks. If the function that returns something to defer actually errors and returns a nil value, go will panic when it tries to add the deferred function to the stack. 

https://play.golang.org/p/1vSonKqczgf

cmd/adminserver.go
```diff
@@ -0,0 +1,46 @@
...
+	Run:          startAdminServer,
+}
+
+func startAdminServer(cmd *cobra.Command, args []string) {
+	var ctx, cancel = context.WithCancel(context.Background())
+	defer cancel()
+
+	logger, err := createLogger()
+	if err != nil {
+		panic(err)
```
jaysonesmith, line 24: What's your reasoning behind panicking here instead of letting cobra handle errors?

quinn-freshly (approved):

cmd/adminserver.go
```diff
@@ -0,0 +1,46 @@
...
+	triggersProjectName := viper.GetString("review-apps-triggers-project-name")
+	if triggersProjectName == "" {
+		panic("need a review apps triggers project name")
+	}
+
+	creds, err := credentials()
+	if err != nil {
+		panic(err)
+	}
+	err = adminserver.Start(ctx, logger, triggersProjectName, creds, viper.GetBool("reviewapps-enabled"), viper.GetString("cluster-default-host"))
```
jaysonesmith, line 38: Could I request a check on whether or not the cluster-default-host value is actually set, please? If it's not the server's `dumbLink` function ends up returning a string of `https://potatoes./` which also gets output to the page as `Available at - https://potatoes./ - if it uses...`

pkg/adminserver/server.go
```diff
@@ -0,0 +1,71 @@
...
+	}.start()
+}
+
+func (s server) start() error {
+	r := mux.NewRouter().PathPrefix("/tuber").Subrouter().StrictSlash(true)
+	for _, handler := range s.handlers {
+		r.HandleFunc(handler.path(), handler.handle)
+	}
+	http.Handle("/", r)
+	return http.ListenAndServe(":3000", nil)
```
jaysonesmith, line 66: Should the port be configurable via env?

pkg/reviewapps/triggers.go
```diff
@@ -72,13 +71,11 @@ func CreateAndRunTrigger(ctx context.Context, logger *zap.Logger, creds []byte,
...
 func deleteReviewAppTrigger(ctx context.Context, creds []byte, project string, reviewAppName string) error {
-	config, err := k8s.GetConfigResource(tuberReviewTriggersConfig, "tuber", "configmap")
+	config, err := k8s.GetConfigResource(TuberReviewTriggersConfig, "tuber", "configmap")
 	if err != nil {
 		return err
 	}
 
-	spew.Dump(config.Data)
-
 	triggerID := config.Data[reviewAppName]
```
jaysonesmith, line 79: linter: `var triggerId should be triggerID`

pkg/adminserver/server.go
```diff
@@ -0,0 +1,71 @@
...
+	ctx    context.Context
+	logger *zap.Logger
+	creds  []byte
+}
+
+func Start(ctx context.Context, logger *zap.Logger, triggersProjectName string, creds []byte, reviewAppsEnabled bool, clusterDefaultHost string) error {
+	var cloudbuildClient *cloudbuild.Service
+	var reviewAppsArgs *createReviewAppsArgs
+	if reviewAppsEnabled {
+		cloudbuildService, err := cloudbuild.NewService(ctx, option.WithCredentialsJSON(creds))
```
jaysonesmith, line 36: If `Start` somehow gets called with a nil or empty byte slice for `creds` while the review apps  enabled is set to true this line will panic.

pkg/adminserver/dashboard.go
jaysonesmith: What if we assigned it(and sourceapps path) to an env var and just allowed anything that needed it to use things that way? It'd allow for updating paths in one spot and also avoid hard coding.

pkg/adminserver/createreviewapp.go
```diff
@@ -0,0 +1,43 @@
...
+		errorParam:   "error",
+		creationPath: "createReviewApp",
+	}
+}
+
+type createReviewApp struct {
+	reviewAppsEnabled     bool
+	creationArgs          *createReviewAppsArgs
+	triggersProjectName   string
+	reviewAppCreatedParam string
```
jaysonesmith, line 19: Unused

pkg/adminserver/dashboard.go
```diff
@@ -0,0 +1,69 @@
...
+	data      struct {
+		AppPath    string
+		SourceApps []SourceApp
+		Error      string
+	}
+}
+
+func (d dashboard) path() string { return d.routePath }
+
+func (d dashboard) setup() handler { return d }
```
jaysonesmith, line 41: Unused

pkg/adminserver/reviewapp.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	reviewApp, err := allReviewApps.FindApp(reviewAppName)
+	if err != nil {
+		data.Error = err.Error()
+		return
+	}
+	data.Branch = reviewApp.Tag
+
+	config, err := k8s.GetConfigResource(reviewapps.TuberReviewTriggersConfig, "tuber", "configmap")
```
jaysonesmith, line 111: Unchecked error

pkg/adminserver/server.go
```diff
@@ -0,0 +1,71 @@
...
+type handler interface {
+	path() string
+	handle(http.ResponseWriter, *http.Request)
+}
+
+type server struct {
+	projectName       string
+	reviewAppsEnabled bool
+	handlers          []handler
+	cloudbuildClient  *cloudbuild.Service
```
jaysonesmith, line 23: Unused

PR #175, opened 2021-03-24
warn instead of explode on bad tag

jkminneti:


Hbbb (approved):

jaysonesmith (approved):
What's the behavior/implications of digest's not having the correct information?

jkminneti:
> What's the behavior/implications of digest's not having the correct information?

@jaysonesmith worst case scenario is that someone messed with their tuber app's cloudbuild yaml and is pushing tags that would never work with tuber - and then misses the fact that tuber isn't deploying them - and then continues merging their way into an n>1 latest to live diff. 

which is basically exactly what happened today - but on changing the gcr repo from `gcr.io` to `us.gcr.io` without telling tuber about it (so they just missed `tuber apps set-repo`'ing in between (which this check would not and did not break on)). they actually made it to n=5 before noticing!

IMO - that's the bigger issue, and the solve is greater transparency about when tuber ACTUALLY deploys things

so tldr: slack notifications.

if people can get used to merging turning quickly into a deployed notification, NOT having one may stand out. 

and otherwise there's just not much we can do in this part of the code. EVERY BUILD from our build project is considered by tuber, and tuber's conventions can't control non-tuber-apps.

PR #178, opened 2021-04-12
admin dash - gin

jkminneti:
sorry quinn

### DIFF
--- a/cmd/adminserver.go
+++ b/cmd/adminserver.go
@@ -1,0 +1,47 @@
+package cmd
+
+import (
+	"context"
+
+	"github.com/freshly/tuber/pkg/adminserver"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+var adminserverCmd = &cobra.Command{
+	SilenceUsage: true,
+	Use:          "adminserver",
+	Short:        "starts the admin http server for review apps and maybe other stuff who knows",
+	Run:          startAdminServer,
+}
+
+func startAdminServer(cmd *cobra.Command, args []string) {
+	var ctx, cancel = context.WithCancel(context.Background())
+	defer cancel()
+
+	logger, err := createLogger()
+	if err != nil {
+		panic(err)
+	}
+
+	defer logger.Sync()
+
+	triggersProjectName := viper.GetString("review-apps-triggers-project-name")
+	if triggersProjectName == "" {
+		panic("need a review apps triggers project name")
+	}
+
+	creds, err := credentials()
+	if err != nil {
+		panic(err)
+	}
+	err = adminserver.Start(ctx, logger, triggersProjectName, creds, viper.GetBool("reviewapps-enabled"), viper.GetString("cluster-default-host"))
+	if err != nil {
+		panic(err)
+	}
+}
+
+func init() {
+	rootCmd.AddCommand(adminserverCmd)
+}
--- a/pkg/adminserver/app.go
+++ b/pkg/adminserver/app.go
@@ -1,0 +1,87 @@
+package adminserver
+
+import (
+	"fmt"
+	"net/http"
+	"sort"
+
+	"github.com/freshly/tuber/pkg/core"
+	"github.com/gin-gonic/gin"
+)
+
+type appResponse struct {
+	Title                  string
+	Error                  string
+	ReviewAppsEnabled      bool
+	Name                   string
+	ReviewAppCreationError string
+	ReviewApps             []appReviewApp
+	Link                   string
+}
+
+type appReviewApp struct {
+	Name   string
+	Branch string
+}
+
+func (s server) app(c *gin.Context) {
+	template := "app.html"
+	appName := c.Param("appName")
+
+	var status = http.StatusOK
+
+	response := &appResponse{
+		Title:                  fmt.Sprintf("Tuber Admin: %s", appName),
+		ReviewAppsEnabled:      s.reviewAppsEnabled,
+		Name:                   appName,
+		ReviewAppCreationError: c.Query("error"),
+	}
+
+	if s.reviewAppsEnabled {
+		reviewApps, err := reviewApps(appName)
+		if err != nil {
+			response.Error = err.Error()
+			c.HTML(http.StatusInternalServerError, template, response)
+			return
+		}
+		response.ReviewApps = reviewApps
+	}
+
+	response.Link = fmt.Sprintf("https://%s.%s/", appName, s.clusterDefaultHost)
+	c.HTML(status, template, response)
+}
+
+func reviewApps(sourceAppName string) ([]appReviewApp, error) {
+	allReviewApps, err := core.TuberReviewApps()
+	if err != nil {
+		return nil, err
+	}
+
+	sourceApps, err := core.TuberSourceApps()
+	if err != nil {
+		return nil, err
+	}
+
+	sourceApp, err := sourceApps.FindApp(sourceAppName)
+	if err != nil {
+		return nil, err
+	}
+
+	var reviewAppsList core.AppList
+	for _, reviewApp := range allReviewApps {
+		if sourceApp.Repo == reviewApp.Repo {
+			reviewAppsList = append(reviewAppsList, reviewApp)
+		}
+	}
+
+	sort.Slice(reviewAppsList, func(i, j int) bool {
+		return reviewAppsList[i].Name < reviewAppsList[j].Name
+	})
+
+	var reviewApps []appReviewApp
+	for _, app := range reviewAppsList {
+		reviewApps = append(reviewApps, appReviewApp{Name: app.Name, Branch: app.Tag})
+	}
+
+	return reviewApps, err
+}
--- a/pkg/adminserver/app.go
+++ b/pkg/adminserver/app.go
@@ -28,8 +28,6 @@ func (s server) app(c *gin.Context) {
 	template := "app.html"
 	appName := c.Param("appName")
 
-	var status = http.StatusOK
-
 	response := &appResponse{
 		Title:                  fmt.Sprintf("Tuber Admin: %s", appName),
 		ReviewAppsEnabled:      s.reviewAppsEnabled,
@@ -48,7 +46,7 @@ func (s server) app(c *gin.Context) {
 	}
 
 	response.Link = fmt.Sprintf("https://%s.%s/", appName, s.clusterDefaultHost)
-	c.HTML(status, template, response)
+	c.HTML(http.StatusOK, template, response)
 }
 
 func reviewApps(sourceAppName string) ([]appReviewApp, error) {
--- a/pkg/adminserver/createreviewapp.go
+++ b/pkg/adminserver/createreviewapp.go
@@ -8,13 +8,12 @@ import (
 )
 
 func (s server) createReviewApp(c *gin.Context) {
-	branch := c.PostForm("branch")
-	appName := c.Param("appName")
-	reviewAppName, err := reviewapps.CreateReviewApp(s.ctx, s.logger, branch, appName, s.creds, s.triggersProjectName)
+	reviewAppName, err := reviewapps.CreateReviewApp(c.Request.Context(), s.logger, c.PostForm("branch"), c.Param("appName"), s.creds, s.triggersProjectName)
 	if err == nil {
 		c.Redirect(http.StatusFound, "reviewapps/"+reviewAppName)
 	} else {
-		errString := "review app creation error: " + err.Error()
-		c.Redirect(http.StatusFound, "?error="+errString)
+		s.logger.Error("review app creation error: " + err.Error())
+		clientError := "review app creation failed"
+		c.Redirect(http.StatusFound, "?error="+clientError)
 	}
 }
--- a/pkg/adminserver/createreviewapp.go
+++ b/pkg/adminserver/createreviewapp.go
@@ -1,0 +1,20 @@
+package adminserver
+
+import (
+	"net/http"
+
+	"github.com/freshly/tuber/pkg/reviewapps"
+	"github.com/gin-gonic/gin"
+)
+
+func (s server) createReviewApp(c *gin.Context) {
+	branch := c.PostForm("branch")
+	appName := c.Param("appName")
+	reviewAppName, err := reviewapps.CreateReviewApp(s.ctx, s.logger, branch, appName, s.creds, s.triggersProjectName)
+	if err == nil {
+		c.Redirect(http.StatusFound, "reviewapps/"+reviewAppName)
+	} else {
+		errString := "review app creation error: " + err.Error()
+		c.Redirect(http.StatusFound, "?error="+errString)
+	}
+}
--- a/pkg/adminserver/dashboard.go
+++ b/pkg/adminserver/dashboard.go
@@ -22,11 +22,12 @@ type sourceApp struct {
 func (s server) dashboard(c *gin.Context) {
 	sourceApps, err := sourceApps()
 	var status = http.StatusOK
-	data := dashboardResponse{Title: "potate"}
+	data := dashboardResponse{Title: "Tuber Dashboard"}
 
 	if err != nil {
 		status = http.StatusInternalServerError
-		data.Error = err.Error()
+		s.logger.Error("error rendering dashboard: " + err.Error())
+		data.Error = "internal error pulling source apps for dashboard"
 	} else {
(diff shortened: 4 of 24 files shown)


pkg/adminserver/reviewapp.go
```diff
@@ -0,0 +1,94 @@
...
+	SourceAppName     string
+	NoSuccessfulBuild bool
+	Builds            []Build
+}
+
+func (s server) reviewApp(c *gin.Context) {
+	template := "reviewApp.html"
+	sourceAppName := c.Param("appName")
+	reviewAppName := c.Param("reviewAppName")
+	var status = http.StatusOK
```
jaysonesmith, line 34: Can we normalize var creation here?

jaysonesmith: It's actually -only- being used on the very last line of this method and every other call to `c.HTML` uses the http type helper directly in the call. Can we actually get rid of this and normalize to the rest of the calls?

pkg/adminserver/reviewapp.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	var successfulBuildExists bool
+	for _, build := range builds {
+		if build.Status == "SUCCESS" {
+			successfulBuildExists = true
+			break
+		}
+	}
+	data.NoSuccessfulBuild = !successfulBuildExists
```
jaysonesmith, line 64: It'd probably be easier to follow some of this by changing the field to successfulBuild setting it to true when a success is found. The template will need to have the if swapped also. As it's written right now, there's a bit of double negative mental gymnastics to follow. 

Additionally, the update would allow us to use the zero value as a meaningful default.

pkg/adminserver/reviewapp.go
```diff
@@ -0,0 +1,94 @@
...
+
+	data.Builds = builds
```
jaysonesmith, line 66: I prefer this up on 56 closer to where it's defined for readability. After that, if you wanted the for loop could also be changed to iterate over `data.Builds` instead. 🤷

pkg/adminserver/reviewapp.go
```diff
@@ -0,0 +1,94 @@
...
+	c.HTML(status, template, data)
+}
+
+func reviewAppBuilds(reviewAppName string, triggersProjectName string, cloudbuildClient *cloudbuild.Service) ([]Build, error) {
+	config, err := k8s.GetConfigResource(reviewapps.TuberReviewTriggersConfig, "tuber", "configmap")
```
jaysonesmith, line 71: Unchecked error.

pkg/adminserver/app.go
```diff
@@ -0,0 +1,87 @@
...
+type appReviewApp struct {
+	Name   string
+	Branch string
+}
+
+func (s server) app(c *gin.Context) {
+	template := "app.html"
+	appName := c.Param("appName")
+
+	var status = http.StatusOK
```
jaysonesmith, line 31: Same thing here about the single use variable and mismatched indirect/direct usage

pkg/adminserver/createreviewapp.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"net/http"
+
+	"github.com/freshly/tuber/pkg/reviewapps"
+	"github.com/gin-gonic/gin"
+)
+
+func (s server) createReviewApp(c *gin.Context) {
+	branch := c.PostForm("branch")
+	appName := c.Param("appName")
```
jaysonesmith, line 12: These are both only used once

pkg/adminserver/dashboard.go
```diff
@@ -0,0 +1,54 @@
...
+
+type sourceApp struct {
+	Name   string
+	Branch string
+}
+
+func (s server) dashboard(c *gin.Context) {
+	sourceApps, err := sourceApps()
+	var status = http.StatusOK
+	data := dashboardResponse{Title: "potate"}
```
jaysonesmith, line 25: Ooooh, I assume this is where the status stuff I commented on earlier originated here?

pkg/adminserver/deletereviewapp.go
```diff
@@ -0,0 +1,20 @@
...
+
+func (s server) deleteReviewApp(c *gin.Context) {
+	sourceAppName := c.Param("appName")
+	reviewAppName := c.Param("reviewAppName")
+	err := reviewapps.DeleteReviewApp(s.ctx, reviewAppName, s.creds, s.triggersProjectName)
+	if err == nil {
+		c.Redirect(http.StatusFound, "/tuber/apps/"+sourceAppName)
+	} else {
+		errString := "review app deletion error: " + err.Error()
+		c.Redirect(http.StatusFound, "/tuber/apps/"+sourceAppName+"?error="+errString)
```
jaysonesmith, line 18: Concatenationception! 😅

jaysonesmith: During your testing did you ever find that the value for errString got out of hand or anything with stuff bubbling up?

pkg/adminserver/reviewapp.go
```diff
@@ -0,0 +1,94 @@
...
+	Name              string
+	Link              string
+	SourceAppName     string
+	NoSuccessfulBuild bool
+	Builds            []Build
+}
+
+func (s server) reviewApp(c *gin.Context) {
+	template := "reviewApp.html"
+	sourceAppName := c.Param("appName")
```
jaysonesmith, line 32: This var is only used once on line 38. Do you feel it would be cleaner to just call `c.Param` directly on 38 instead of creating a single use variable?

pkg/adminserver/templates/reviewApp.html
```diff
@@ -0,0 +1,20 @@
...
+		<br>
+		<p>no successful build exists yet, so the app will not work.</p>
+  {{end}}
+	<h2>Builds</h2>
+	<table>{{range $build := .Builds}}
+			<tr>
+				<td><a href="{{$build.Link}}">{{$build.StartTime}}</a></td>
+				<td>{{$build.Status}}</td>
+			</tr>
+      {{end}}</table>
```
jaysonesmith, line 19: Not sure if the placement of the table elements this was for output whitespace, but you can do this too (iirc):

```
<table>
      {{- range $build := .Builds -}}
            <tr>
                  <td><a href="{{$build.Link}}">{{$build.StartTime}}</a></td>
                  <td>{{$build.Status}}</td>
            </tr>
      {{end -}}
</table>
  ```

The space hyphen indicates to trim whitespace from between the template element and the html elemet.

**Review by jaysonesmith — APPROVED** [review]

File: pkg/adminserver/server.go [hunk 11]
```diff
@@ -0,0 +1,62 @@
...
+	"google.golang.org/api/option"
+)
+
+type server struct {
+	projectName         string
+	reviewAppsEnabled   bool
+	cloudbuildClient    *cloudbuild.Service
+	clusterDefaultHost  string
+	triggersProjectName string
+	ctx                 context.Context
```
Hbbb, line 18: This is an anti-pattern according to the official Go documentation. https://blog.golang.org/context-and-structs.

Context is explicitly defined as a request-scoped argument that should be passed in to each handler.

pkg/adminserver/server.go
```diff
@@ -0,0 +1,62 @@
...
+		cloudbuildClient:    cloudbuildClient,
+		clusterDefaultHost:  clusterDefaultHost,
+		triggersProjectName: triggersProjectName,
+		ctx:                 ctx,
+		logger:              logger,
+		creds:               creds,
+	}.start()
+}
+
+func (s server) start() error {
```
Hbbb, line 46: Why not use a pointer receiver here? I realize you're not modifying anyting on `server`, but it is a fairly large struct to be copying all the time. This applies to all of the methods on server.

pkg/adminserver/deletereviewapp.go
```diff
@@ -0,0 +1,20 @@
...
+	"net/http"
+
+	"github.com/freshly/tuber/pkg/reviewapps"
+	"github.com/gin-gonic/gin"
+)
+
+func (s server) deleteReviewApp(c *gin.Context) {
+	sourceAppName := c.Param("appName")
+	reviewAppName := c.Param("reviewAppName")
+	err := reviewapps.DeleteReviewApp(s.ctx, reviewAppName, s.creds, s.triggersProjectName)
```
Hbbb, line 13: Expanding on my other comment about storing context on `server`, shouldn't you be passing the `c *gin.Context` around? Or using the underlying context `c.Request.Context()` and passing that around? I'm fairly certain we shouldn't be storing context on a struct.

pkg/adminserver/dashboard.go
Hbbb: Were you meaning to change the title here?

pkg/adminserver/dashboard.go
```diff
@@ -0,0 +1,54 @@
...
+
+	if err != nil {
```
Hbbb, line 27: I think it makes sense to wrap this error with a more context-appropriate message. Right now, if `core.TuberSourceApps()` fails, the error message it will return will come all the way from `os.Exec`.

pkg/adminserver/createreviewapp.go
```diff
@@ -0,0 +1,20 @@
...
+	reviewAppName, err := reviewapps.CreateReviewApp(s.ctx, s.logger, branch, appName, s.creds, s.triggersProjectName)
```
Hbbb, line 13: Maybe I should just leave this as the final comment of the review because I think it applies to a lot of the `reviewapps` functions. I think it makes sense to wrap errors inside `reviewapps.CreateReviewApp` with a more contextually relevant string. I don't think it makes sense to render `kubectl` errors to the UI.

pkg/adminserver/reviewapp.go
```diff
@@ -0,0 +1,94 @@
...
+
+func reviewAppBuilds(reviewAppName string, triggersProjectName string, cloudbuildClient *cloudbuild.Service) ([]Build, error) {
+	config, err := k8s.GetConfigResource(reviewapps.TuberReviewTriggersConfig, "tuber", "configmap")
+	triggerId := config.Data[reviewAppName]
+	if triggerId == "" {
+		return nil, fmt.Errorf("trigger is untracked or it doesnt exist")
+	}
+
+	buildsResponse, err := cloudbuild.NewProjectsBuildsService(cloudbuildClient).List(triggersProjectName).Filter(fmt.Sprintf(`trigger_id="%s"`, triggerId)).Do()
+	if err != nil {
```
Hbbb, line 78: Maybe wrap this error? Cloud Build service errors seem like they'd be scary and hard to decipher.

pkg/adminserver/reviewapp.go
```diff
@@ -0,0 +1,94 @@
...
+		return nil, err
+	}
+
+	var builds []Build
+	for _, build := range buildsResponse.Builds {
+		var startTime string
+		if build.StartTime != "" {
+			parsed, timeErr := time.Parse(time.RFC3339, build.StartTime)
```
Hbbb, line 86: If there's an error parsing `build.StartTime`, won't `parsed` be nil?

Hbbb (changes requested):
A few small things

pkg/adminserver/createreviewapp.go
jkminneti: i sorta liked how all of the stuff we're pulling from the request is at the top, but, looking at it now, i dont mind the alternative

pkg/adminserver/dashboard.go
jkminneti: > Were you meaning to change the title here?

potate

pkg/adminserver/deletereviewapp.go
jkminneti: thought was while this feature is in infancy, it'd make things easier to debug. given we're jumping right into client usage here, wrapping all de errors and logging instead

pkg/adminserver/templates/reviewApp.html
jkminneti: havent really landed on how i like to see go templated html and goland does weird stuff with it anyway.

going with this for now because i definitely wasnt trying to do anything with whitespace:
```go
	<table>
			{{range $build := .Builds}}
			<tr>
				<td><a href="{{$build.Link}}">{{$build.StartTime}}</a></td>
				<td>{{$build.Status}}</td>
			</tr>
      {{end}}
	</table>
```

cmd/adminserver.go
```diff
@@ -0,0 +1,47 @@
...
+	triggersProjectName := viper.GetString("review-apps-triggers-project-name")
+	if triggersProjectName == "" {
+		panic("need a review apps triggers project name")
+	}
+
+	creds, err := credentials()
+	if err != nil {
+		panic(err)
+	}
+	err = adminserver.Start(ctx, logger, triggersProjectName, creds, viper.GetBool("reviewapps-enabled"), viper.GetString("cluster-default-host"))
```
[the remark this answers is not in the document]
jkminneti: it's not necessary for startup

pkg/reviewapps/triggers.go
```diff
@@ -12,12 +12,12 @@ import (
 	"google.golang.org/api/option"
 )
 
-const tuberReposConfig = "tuber-repos"
-const tuberReviewTriggersConfig = "tuber-review-triggers"
+const TuberReposConfig = "tuber-repos"
+const TuberReviewTriggersConfig = "tuber-review-triggers"
```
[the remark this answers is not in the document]
jkminneti: they're used in this pr:
```
func reviewAppBuilds(reviewAppName string, triggersProjectName string, cloudbuildClient *cloudbuild.Service) ([]Build, error) {
	config, err := k8s.GetConfigResource(reviewapps.TuberReviewTriggersConfig, "tuber", "configmap")
```

pkg/adminserver/server.go
jkminneti: _does_ it get copied?

like if we tell gin to use 
```
tuber.GET("/", s.dashboard)
```
does passing that function imply a new copy is made every time that function is run?

pkg/adminserver/createreviewapp.go
jkminneti: posting this here too

thought was while this feature is in infancy, it'd make things easier to debug. given we're jumping right into client usage here, wrapping all de errors and logging instead

pkg/adminserver/reviewapp.go
jkminneti: nope, just `Time{}` (so `January 1, year 1, 00:00:00.000000000 UTC`)

pkg/adminserver/createreviewapp.go
```diff
@@ -8,13 +8,12 @@ import (
...
-	appName := c.Param("appName")
-	reviewAppName, err := reviewapps.CreateReviewApp(s.ctx, s.logger, branch, appName, s.creds, s.triggersProjectName)
+	reviewAppName, err := reviewapps.CreateReviewApp(c.Request.Context(), s.logger, c.PostForm("branch"), c.Param("appName"), s.creds, s.triggersProjectName)
 	if err == nil {
 		c.Redirect(http.StatusFound, "reviewapps/"+reviewAppName)
 	} else {
-		errString := "review app creation error: " + err.Error()
-		c.Redirect(http.StatusFound, "?error="+errString)
+		s.logger.Error("review app creation error: " + err.Error())
+		clientError := "review app creation failed"
```
jaysonesmith, line 16: c6 - Did you mean to declare this to be used by both line 15 and 17?

jaysonesmith (approved):

pkg/adminserver/server.go
jaysonesmith: https://tour.golang.org/methods/8

cmd/adminserver.go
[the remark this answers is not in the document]
Hbbb: Fair enough!

pkg/reviewapps/triggers.go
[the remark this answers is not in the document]
Hbbb: 👍

pkg/adminserver/server.go
Hbbb: Yea it gets copied on each invocation. I don't know that this is a _huge_ issue, it's more of a best practices thing.

pkg/adminserver/reviewapp.go
Hbbb: Alright, no big deal.

Hbbb (approved):

PR #180, opened 2021-04-17
Check for env vars in config when passing tuber a file to set

jaysonesmith:
This enables using local config files that point to env vars instead of saving tokens and stuff to the file itself

```
CARDINAL_SLACK_TOKEN: ${SLACK_OAUTH_TOKEN}
CARDINAL_SLACK_CHANNEL: ${SLACK_CARDINAL_CHANNEL}
CARDINAL_SENTRY_DSN: ${CARDINAL_SENTRY_DSN}
```

### DIFF
--- a/pkg/k8s/secrets.go
+++ b/pkg/k8s/secrets.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 
 	"github.com/goccy/go-yaml"
 )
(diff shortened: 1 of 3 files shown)


jkminneti:
seems like a lot of local work for the client to produce a file customized with the implied formatting just to spare generating a temporary file with the values hardcoded

why not lean into dotenv instead and have `env list` and `env file` optionally return and implicitly accept that format, such that local usage from an app's working directory "just works", for running with the state of a remote app or for updating that state

Hbbb (approved):

jaysonesmith:
> seems like a lot of local work for the client to produce a file customized with the implied formatting just to spare generating a temporary file with the values hardcoded
> 
> why not lean into dotenv instead and have `env list` and `env file` optionally return and implicitly accept that format, such that local usage from an app's working directory "just works", for running with the state of a remote app or for updating that state

@jkminneti 

From looking at godotenv's documentation, it appears that you still have to have the actual value in the file which is ultimately what I don't like. If I've already got them stored in my zshrc file or somewhere else, I'd prefer to keep them there instead of copy/pasting them to another location.

I also don't personally like the idea of having keys in the project itself. In most situations we'd likely be good about gitignore'ing files, but what if we're not, ya know?

I'm also not sure what about it looks like a lot of work as you mentioned. Can you clarify? This code is also backwards compatible, so you or whoever can continue creating files as you always have.

jkminneti:
> > seems like a lot of local work for the client to produce a file customized with the implied formatting just to spare generating a temporary file with the values hardcoded
> > why not lean into dotenv instead and have `env list` and `env file` optionally return and implicitly accept that format, such that local usage from an app's working directory "just works", for running with the state of a remote app or for updating that state
> 
> @jkminneti
> 
> From looking at godotenv's documentation, it appears that you still have to have the actual value in the file which is ultimately what I don't like. If I've already got them stored in my zshrc file or somewhere else, I'd prefer to keep them there instead of copy/pasting them to another location.
> 
> I also don't personally like the idea of having keys in the project itself. In most situations we'd likely be good about gitignore'ing files, but what if we're not, ya know?
> 
> I'm also not sure what about it looks like a lot of work as you mentioned. Can you clarify? This code is also backwards compatible, so you or whoever can continue creating files as you always have.

i see. i usually don't rely on (bash_profile) for anything that `.env` can handle, cus i find it easier to forget stuff is globally set.

but if you're worried about having non-git-ignored files in your dir, wouldn't this _have_ to be an additional file next to your `.env`, if you wanted to use `.env` to run the app? that's what i meant by extra work - I just can't imagine using it in conjunction with `.env`

my suggestion would make `.env` usage better/easier, this pr would make NOT using `.env` easier. 

so i guess long story short is there's no harm in it, even if i'd never use it lol

pkg/k8s/secrets.go
```diff
@@ -64,38 +65,53 @@ func GetSecret(namespace string, secretName string) (*ConfigResource, error) {
...
+	// look for a value of ${SOME_LOCAL_ENV_VAR}
+	// works with env vars that have hyphens or no separators
+	var envVar = regexp.MustCompile(`\$\{(.*)\}`)
+
+	var out = make(map[string]string, len(data))
+	for k, v := range data {
+		val := fmt.Sprint(v)
+		found := envVar.FindStringSubmatch(val)
+		if len(found) > 1 {
+			val = os.Getenv(found[1])
```
jkminneti, line 109: i suppose this would break down a bit if you were actually TRYING to set an env var to the literal value `"${asdf}"` - not sure how realistic that is though. 

if getenv `== ""`, what if we left the value alone? would that be any worse than empty string?

there's also non-alpha encrypted vars/passwords to consider, if you happened to get a `${` and a `}` down the line

PR #188, opened 2021-05-14
just flip everything

jkminneti:
it fucking works


data/tuberapps/role.go
```diff
@@ -4,5 +4,5 @@ package data
...
+	Contents: []byte{ 107,105,110,100,58,32,82,111,108,101,13,10,97,112,105,86,101,114,115,105,111,110,58,32,114,98,97,99,46,97,117,116,104,111,114,105,122,97,116,105,111,110,46,107,56,115,46,105,111,47,118,49,98,101,116,97,49,13,10,109,101,116,97,100,97,116,97,58,13,10,32,32,110,97,109,101,58,32,116,117,98,101,114,45,97,100,109,105,110,13,10,32,32,110,97,109,101,115,112,97,99,101,58,32,123,123,46,110,97,109,101,115,112,97,99,101,125,125,13,10,114,117,108,101,115,58,13,10,32,32,45,32,97,112,105,71,114,111,117,112,115,58,13,10,32,32,32,32,45,32,39,42,39,13,10,32,32,32,32,114,101,115,111,117,114,99,101,115,58,13,10,32,32,32,32,32,32,45,32,39,42,39,13,10,32,32,32,32,118,101,114,98,115,58,13,10,32,32,32,32,32,32,45,32,39,42,39, },
```
quinn-freshly, line 7: ok so. did you know that Windows line endings are actually _two characters_ compared to linux line endings. one of the characters is shared between them, called LF, and in UTF8 decimal byte representation, that is 10. The other character, coming before the LF, and only used by windows, CR for carriage return the byte value in decimal for this is 13. So, if you were to regenerate a byte representation of a text file that had been converted FROM linux TO windows, it would add number 13 in front of number 10 in a bunch of places.

jkminneti: i did know that, and im running generate from wsl every time... weird

quinn-freshly (approved):
Can i merge this

PR #190, opened 2021-05-16
i just wrote an orm

jkminneti:
on a saturday

open questions:
1. is this stupid

[pkg/db/db.go](https://github.com/Freshly/tuber/blob/i-just-wrote-an-orm/pkg/db/db.go) < orm
[graph/model/models.go](https://github.com/Freshly/tuber/blob/i-just-wrote-an-orm/graph/model/models.go) < interface for tuberapps
[pkg/core/apps.go](https://github.com/Freshly/tuber/blob/i-just-wrote-an-orm/pkg/core/apps.go) < query layer for tuberapps


pkg/db/db.go
```diff
@@ -0,0 +1,319 @@
...
+			var match = true
+			for queryKey, queryVal := range normalizedQueryValues {
+				storedVal := rootbentryb.Get([]byte(queryKey))
+				if string(storedVal) != queryVal {
+					match = false
+					break
+				}
+			}
+			if match {
+				marshalled = append(marshalled, queryResult{key: string(k), marshalled: rootbentryb.Get([]byte(marshalledKey))})
```
quinn-freshly, line 192:
```suggestion
				marshalled = append(marshalled, rootbentryb.Get([]byte(marshalledKey))...)
				marshalled = append(marshalled, []byte(",")...)
				
```

and then sometime later:

```
marshalled = append(marshalled, []byte("]")...)
```

pkg/db/db.go
```diff
@@ -0,0 +1,319 @@
...
+			return nil
+		})
+
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	var results []Model
```
quinn-freshly, line 203: `results` would have been passed in, and we would do a single unmarshal into it. done!!

graph/model/models.go
```diff
@@ -0,0 +1,38 @@
...
+	return "apps"
+}
+
+func (t TuberApp) DBKey() string {
+	return t.Name
+}
+
+func (t TuberApp) DBMarshal() ([]byte, error) {
+	return json.Marshal(t)
+}
```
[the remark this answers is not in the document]
jkminneti: hmmm whole point of this was to allow anything to be marshalled in any way, including any logic we could want

one model stores as yaml and composes a couple index vals from the base struct? sure!
another's base model needs a lighter interface and you want a composite struct to handle these functions? sure!
another stores as json and needs nothing on its way? sure!

by leaving it up to the model, you leave it up to the client, which keeps the orm _super frickin lightweight_

pkg/core/apps.go
```diff
@@ -0,0 +1,110 @@
+package core
```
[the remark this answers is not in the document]
jkminneti: in some ways yes, but I still want these functions to exist, as a form of canonical common queries

better question is if this should expose the underlying `db.DB`, as we're passing around `core.DB`s that _only_ have these methods currently. keeps it nice and clearly defined, but you have to really like the idea of canonical common queries

pkg/db/db.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+
+	return &DB{db: db}, nil
+}
+
+func (d *DB) Close() {
+	d.db.Close()
+}
+
+const marshalledKey = "marshalled"
```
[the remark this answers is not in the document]
jkminneti: yep good point. i can do `__obj` if you want lol

pkg/db/db.go
jkminneti: oh i dont like that at alllllll lol - took me forever to figure out what you were suggesting; i think clarity should win out

pkg/db/db.go
jkminneti: we're working with interfaces though, don't think that'll work as an unmarshal target

pkg/core/apps.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	return assertAll(r)
+}
+
+func (d *DB) AppExists(appName string) bool {
+	return d.db.Exists(model.TuberApp{}, appName)
+}
+
+func (d *DB) AppsForTag(tag string) ([]*model.TuberApp, error) {
+	r, err := d.db.Get(model.TuberApp{}, db.Q().String("imageTag", tag))
```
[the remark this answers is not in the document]
jkminneti: unmarshalling does feel like something the orm should be equipped to do though. 

but beyond that, with this we lose the ability to validate queries - invalid queries could be very hard to detect.

but you're right though - with this there is no need for an interface in between, interesting

graph/model/models.go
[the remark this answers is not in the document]
quinn-freshly: @jkminneti  i'd rather default to json and refactor later if we end up needing it. the refactor would be easy, and we can start with something super minimal for the scope of this initial PR.

pkg/db/db.go
quinn-freshly: @jkminneti you're right that its heinous. but it would make the outside appear more idiomatic. You also would only ever do a single unmarshal, not sure if there any performance considerations there (not that I think we should be considering performance at this stage)

quinn (approved):
only feedback at this point is cosmetic (the method names of the Model interface)

cmd/bolt.go
```diff
@@ -112,10 +114,13 @@ func pullLocalDB(db *core.DB) error {
 }
 
 func cloudrepo(a *model.TuberApp, data map[string]string) (string, error) {
-	repo, err := core.RepoFromTag(a.ImageTag)
+	sourceAppTagGCRRef, err := name.ParseReference(a.ImageTag)
 	if err != nil {
 		return "", err
 	}
+
+	repo := strings.Split(sourceAppTagGCRRef.Name(), ":")[0]
```
jkminneti, line 122: what nooooooooo

cmd/bolt.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/freshly/tuber/graph/model"
 	"github.com/freshly/tuber/pkg/core"
 	"github.com/freshly/tuber/pkg/k8s"
+	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
```
jaysonesmith, line 14: Off the bat, my first bit of feedback would be that I wouldn't call this file `bolt` but rather something that more generally describes the thing being worked with. IMO, something like `database` would be what I'd use.

For the command, it's a similar thing: `tuber bolter` feels real odd to me.  I'd likely do something like a base command of `database` or `db` and then have a sub-command off of it as `pull-remote`. To run it, it'd then be `tuber db pull-remote` which is directly communicative to what's going on. As a someone without context, I wouldn't know what to expect with a command named `bolter`.

Both things above would also make it so that we're not forever locked into bolt as an option & doesn't expose implementation details to the user. Theoretically, this particular command could also be written with a generic database interface allowing other dbs to be wired in.

cmd/bolt.go
```diff
@@ -21,15 +22,15 @@ var bolterCmd = &cobra.Command{
...
 		return err
 	}
+	defer db.Close()
 
 	return pullLocalDB(db)
 }
 
-func pullLocalDB(db *core.Data) error {
+func pullLocalDB(db *core.DB) error {
 	fmt.Println("pulling db from configmaps, takes a sec")
```
jaysonesmith, line 34: No desire for a logger?

cmd/bolt.go
```diff
@@ -68,7 +69,10 @@ func pullLocalDB(db *core.Data) error {
 			Vars:         []*model.Tuple{},
```
jaysonesmith, line 69: I can't comment on the line itself, but line 52 and the check on line 54 seems unneeded. Nothing in the rest of the function has the ability to change `reviewAppsEnabled` so the check is redundant and the other use of the var on line 85 could be set to `true` on its own.

cmd/bolt.go
jkminneti: oh idea was that this command exists for exactly 1 commit on each of develop and master

once the running service has the data locally and a graphql server prepared to offer all said data, the next update kills this command for good

cmd/bolt.go
jkminneti: same deal, this command is 1000% temporary

pkg/db/db.go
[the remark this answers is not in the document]
jaysonesmith: Either that or I've seen something like prefixing with `xx` I think?

cmd/helpers.go
```diff
@@ -38,23 +38,11 @@ func db() (*core.Data, error) {
 	} else {
 		path = "/etc/tuber-bolt/db"
 	}
-	db, err := bolt.Open(path, 0666, nil)
+	database, err := tuberbolt.NewDefaultDB(path, model.TuberApp{}.DBRoot())
```
jaysonesmith, line 41: IMO, if we're calling a function that returns a default, it should take no arguments. The setup that `db()` does before calling this is required, it'd make more sense to be done within `NewDefaultDB`.

cmd/helpers.go
```diff
@@ -11,9 +11,9 @@ import (
 	"github.com/freshly/tuber/graph/client"
 	"github.com/freshly/tuber/graph/model"
 	"github.com/freshly/tuber/pkg/core"
+	tuberbolt "github.com/freshly/tuber/pkg/db"
```
jaysonesmith, line 14: Similar comment about `tuberbolt` as I made around naming the other file `bolt`.

graph/model/models.go
```diff
@@ -0,0 +1,38 @@
...
+	return map[string]string{
+			"name":          t.Name,
+			"imageTag":      t.ImageTag,
+			"sourceAppName": t.SourceAppName,
+		}, map[string]bool{
+			"reviewApp": t.ReviewApp,
+		}, map[string]int{}
+}
+
+func (t TuberApp) DBRoot() string {
```
jaysonesmith, line 19: I'm not digging the `DB` prefix on these methods. While I'm admittedly not familiar with all of this code just yet, they also don't seem communicative as to what they return without having additional context. I think if possible I'd prefer to see some sort of db struct embedded into the app struct so that we could do something like `app.db.key` or something?

graph/model/models.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+func (t TuberApp) DBIndexes() (map[string]string, map[string]bool, map[string]int) {
+	return map[string]string{
+			"name":          t.Name,
+			"imageTag":      t.ImageTag,
+			"sourceAppName": t.SourceAppName,
+		}, map[string]bool{
+			"reviewApp": t.ReviewApp,
+		}, map[string]int{}
```
jaysonesmith, line 16: This is super hard to read and understand easily for me. Could we return a struct with these types in them so that we can get some sort of naming associated to them?

pkg/db/db.go
```diff
@@ -0,0 +1,313 @@
...
+	if key == "" {
+		return fmt.Errorf("save failed, model key nil")
+	}
+
+	strings, bools, ints := m.DBIndexes()
+	var err error
+
+	for k, v := range bools {
+		if strings[k] != "" {
+			return fmt.Errorf("model misconfigured, index keys overlap across type")
```
jaysonesmith, line 86: Could we expand these to add info around what types are being overlapped?

pkg/db/db.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+
+	marshalled, err := m.DBMarshal()
+	if err != nil {
+		return err
+	}
+
+	err = d.db.Update(func(tx *bolt.Tx) error {
+		rootb := tx.Bucket([]byte(m.DBRoot()))
+		mb, err := rootb.CreateBucketIfNotExists([]byte(m.DBKey()))
```
jaysonesmith, line 105: There's no other use of `root` here so appending the `b` to denote it's of byte isn't needed, plus we can see that its of type byte based on the code.

jaysonesmith, line 105: Is there any significance in naming this `rootb` instead of just `b` or `bkt`? Yes, I see that you're calling `DBRoot` from the tuber app, but there's no other bucket reference here and I'm not immediately seeing that specifying it as root provides additional context.

pkg/db/db.go
```diff
@@ -0,0 +1,313 @@
...
+
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func Q() Query {
```
jaysonesmith, line 128: `NewQueryValues` instead? When I see query, I feel like it should be the actual thing that is run to do the query (aka query template & values together)

There's a bigger refactor that I've got a little progress on, but it's much lower priority compared to everything else.

PR #195, opened 2021-05-26
Compile assets

quinn-freshly:


Dockerfile
```diff
@@ -1,6 +1,14 @@
 FROM bitnami/kubectl:1.15-ol-7
```
jaysonesmith, line 1: Fwiw, I think when I was working on Cardinal's Dockerfile I ended up going with the following method to get kubectl installed as it was faster and more lightweight than the bitnami image.

```
# download kubectl
RUN apk update && apk add curl
RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.18.1/bin/linux/amd64/kubectl
RUN chmod u+x kubectl && mv kubectl /bin/kubectl
```

PR #204, opened 2021-06-10
POPS-1111 Remove Env Dependency

Hbbb:
* Removes dependency on env vars in CLI
* Refactors tuber config into `config` package


quinn-freshly (approved):

pkg/config/main.go
```diff
@@ -0,0 +1,105 @@
+package config
```
jkminneti, line 1: can we rename this file `config.go` for my sanity?

cmd/config.go
```diff
@@ -14,7 +15,7 @@ var configCmd = &cobra.Command{
 	Use:          "config",
 	Short:        "open local tuber config in your default editor",
 	Args:         cobra.NoArgs,
-	RunE:         config,
+	RunE:         writeConfig,
```
jkminneti, line 18: it's more like `openConfig` but obviously pedantic c6 or whatever

quinn-freshly (changes requested):
seems to have build errors

pkg/config/main.go
Hbbb: Absolutely.

PR #207, opened 2021-06-12
bunch of stuff that was on staging

jkminneti:
let's set staging to main upon merging this

- debug mode is an env var, dont set it on the deployment
- fix deploy command. it now makes a local empty database, and uses a (now uncommented) graphql query to pull a single app and saves it.
- tell sentry to pick up panics
- monitor prerelease pods for ANY completed container, so that the sql auth proxy container doesn't hold it up indefinitely
- FIX REMOVED RESOURCE CLEANUP
- fix rollout monitoring for resources with monitor urls
- better error handling within gcr
- like, several panics fixed


pkg/core/releaser.go
```diff
@@ -537,8 +540,9 @@ func (r releaser) rollbackResource(applied appResource, cached appResource) erro
 	return nil
 }
 
-func (r releaser) reconcileState(stateBeforeApply []appResource, appliedWorkloads []appResource, appliedConfigs []appResource) error {
+func (r releaser) reconcileState(stateBeforeApply []appResource, appliedWorkloads []appResource, appliedConfigs []appResource, appliedPostreleaseResources []appResource) error {
 	var appliedResources appResources = append(appliedWorkloads, appliedConfigs...)
+	appliedResources = append(appliedWorkloads, appliedPostreleaseResources...)
```
jkminneti, line 545:
```suggestion
	appliedResources = append(appResources, appliedPostreleaseResources...)
```

PR #209, opened 2021-06-13
move frontend to separate deployment

quinn-freshly:


.tuber/virtualservice.yaml
```diff
@@ -13,6 +13,14 @@ spec:
             host: tuber
             port:
               number: 3000
+      match:
+        - uri:
+            prefix: /tuber/graphql
+    - route:
+        - destination:
+            host: tuber-frontend.tuber-frontend.svc.cluster.local
```
jkminneti, line 21: new virtualservice inside of tuber-frontend should handle this

with `{{.tuberAppName}}`'s all the way down

question then becomes how to istio it. can we exact match on `/tuber/graphql` in tuber-frontend, and prefix match on `/tuber`, and have it just work?

or will we have to exact match on `/tuber`, and prefix match on `/tuber/apps` etc

either way should be doable

why? FRONTEND REVIEW APPS 🎉

cloudbuild.yaml
```diff
@@ -0,0 +1,46 @@
+steps:
```
jkminneti, line 1: should split into two cloudbuild yamls, one for frontend one for backend, _both using kaniko..._, and with file filters to match (assuming, hopefully safely, we can segment frontend by a toplevel dir)

pkg/adminserver/web/.tuber/deployment.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: tuber-frontend
```
jkminneti, line 4: if we do go the frontend review apps route, this should all be `{{.tuberAppName}}`'d (same for the service)

pkg/iap/auth.go
```diff
@@ -32,7 +31,6 @@ func init() {
 }
 
 func RefreshTokenExists() bool {
-	spew.Dump(refreshTokenFile)
```
jkminneti, line 35: 🙏 🙏 🙏 🔪

.tuber/persistentvolumeclaim.yaml
```diff
@@ -2,7 +2,6 @@ apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
   name: tuber
-  namespace: tuber
```
jkminneti, line 5: these were there to _explicitly disallow review apps of tuber_

tuber now has a `TuberApp.ReviewAppsConfig.Enabled = false` but stilllllll idk up to you.

jkminneti (changes requested):

.tuber/virtualservice.yaml
jkminneti: actually you probably use relative routing to get to `/tuber/graphql`

so if a review app is running on `/tuber-pops-123`, we'll need another route to define `/tuber-pops-123/graphql`

which means `tuber` doesn't even need a virtualservice!

cloudbuild.yaml
quinn-freshly: idk how 2 kaniko

.tuber/persistentvolumeclaim.yaml
quinn-freshly: ehh, lets just use the setting now that we have it.

jkminneti (approved):
fixes the route

PR #214, opened 2021-06-14
iap client id and also stop panicking

jkminneti:


pkg/iap/auth.go
```diff
@@ -62,17 +67,21 @@ func CreateRefreshToken() error {
...
+	c, err := newOAuthConfig()
+	if err != nil {
+		return "", err
+	}
+
 	v := url.Values{
 		"grant_type":    {"refresh_token"},
 		"refresh_token": {string(refreshToken)},
-		"audience":      {viper.GetString("iap-client-id")},
+		"audience":      {IAPClientID},
```
jkminneti, line 84: that's the actual fix here lol

PR #217, opened 2021-06-15
fix nil pointer

quinn-freshly:


jkminneti:
weirdly enough i disagree with this change. 

`config.Load()` returns an error. to me that's an implied contract that the "non-error" result won't be nil, especially for something that really shouldn't be nil.

and if there was a case where it COULD be nil, WITHOUT an error, that should be fixed in the underlying function

up to you though, this certainly doesn't hurt anything.

quinn-freshly:
it happened to me on main so

jkminneti:
sorry i took one quick glance and thought you were doing `cnf == nil` not `cnf.Auth == nil`

PR #218, opened 2021-06-15
restore set branch cmd

jkminneti:


quinn-freshly (approved):
👍👍 but would’ve been cooler without the churn to set image tag but I get it.

jkminneti:
> 👍👍 but would’ve been cooler without the churn to set image tag but I get it.

that's why inline funcs in cobra commands should be illegal

PR #231, opened 2021-06-20
the final dan ism

jkminneti:
```go
viper.GetString("some-string") -> viper.GetString("TUBER_SOME_STRING")
```
what a destructive bit of "convenience"


main.go
```diff
@@ -4,11 +4,9 @@ import (
 	"log"
 
 	"github.com/freshly/tuber/cmd"
-	"github.com/spf13/viper"
 )
 
 func main() {
-	viper.SetDefault("prefix", "/tuber")
```
quinn-freshly, line 11: this is necessary

quinn-freshly (changes requested):

pkg/adminserver/server.go
```diff
@@ -79,17 +81,18 @@ func localDevServer(res http.ResponseWriter, req *http.Request) {
...
+	return fmt.Sprintf("%s%s", serverPrefix, route)
 }
 
 func (s server) start() error {
+	viper.SetDefault("TUBER_ADMINSERVER_PREFIX", "/tuber")
 	mux := http.NewServeMux()
-	mux.HandleFunc(prefix("/graphql/playground"), playground.Handler("GraphQL playground", prefix("/graphql")))
-	mux.Handle(prefix("/graphql"), graph.Handler(s.db, s.processor, s.logger, s.creds, s.triggersProjectName, s.clusterName, s.clusterRegion))
+	mux.HandleFunc(prefixRoute("/graphql/playground", s.prefix), playground.Handler("GraphQL playground", prefixRoute("/graphql", s.prefix)))
+	mux.Handle(prefixRoute("/graphql", s.prefix), graph.Handler(s.db, s.processor, s.logger, s.creds, s.triggersProjectName, s.clusterName, s.clusterRegion))
```
quinn-freshly, line 92: :(

quinn-freshly (changes requested):

pkg/adminserver/server.go
jkminneti: ?

quinn-freshly: @jkminneti it was so much cleaner before 😢

jkminneti: @quinn-freshly i wanna keep viper (or indeed any env getting) out in the `cmd` package. `pkg` stuff should have explicit inputs imo

we have, afaik, kept that up in all cases but the debug mode check within kubectl.go

quinn-freshly (approved):

PR #233, opened 2021-06-21
Add error update to main

jaysonesmith:
quick add, I think it's good?


quinn-freshly (approved):

jkminneti (approved):
aww i wanted to see how many things that function could return

PR #251, opened 2021-07-22
POPS-1149 Add build statuses to dashboard

Hbbb:
Adds Cloud Build statuses to the dashboard

qa-replicated_-_staging


graph/schema.resolvers.go
```diff
@@ -680,6 +681,15 @@ func (r *tuberAppResolver) Env(ctx context.Context, obj *model.TuberApp) ([]*mod
 	return list, nil
 }
 
+func (r *tuberAppResolver) CloudBuildStatuses(ctx context.Context, obj *model.TuberApp) ([]*model.Build, error) {
+	builds, err := builds.FindByApp(obj, "freshly-docker")
```
jkminneti, line 685: needs to be passed in from the top

jkminneti (changes requested):

graph/schema.resolvers.go
Hbbb: Yep. I noticed that the resolver's `projectName` is actually the server's `triggersProjectName`. The `projectName` field on the server struct is redundant.

jkminneti (approved):

PR #252, opened 2021-07-27
Add JSON and .env output formats to `env list`

Hbbb:


cmd/env.go
```diff
@@ -212,12 +214,26 @@ func envList(cmd *cobra.Command, args []string) error {
...
+	case "env":
+		for k, v := range m {
+			output = append(output, []byte(fmt.Sprintf("%s = \"%s\"\n", k, v))...)
+		}
+	case "json":
+		output, err = json.Marshal(m)
+		if err != nil {
+			return err
+		}
+	default:
```
jkminneti, line 228: it'd be nicer to offer a `yaml`, as the default value of the _flag_, add that case to the switch, and then default to returning an error explaining the options

jkminneti (approved):

cmd/env.go
jkminneti: plus having the options in the help text

PR #254, opened 2021-08-04
user impersonation

jkminneti:
The point of this PR is to offer to our graphql resolvers this function:
```
func GetAccessToken(ctx context.Context) (string, error) {}
```
any endpoint that needs a `kubectl auth can-i` or full user impersonation can rely on that return.

a proof of that is in the dashboard homepage - it does nothing with the output of the can-i, but the proof is in the logs.

-----

This is still an incomplete first step.

I couldn't (or wouldn't) get nextjs to follow server redirects (server sends 301 and a target, and nextjs is like "cool but like where's the data i asked for")

so you have to go to, in the deployed case of internal, `https://admin.freshlyhq.com/tuber/login` to get yourself a cookie or two

But beyond that (easily fixed once supported), this works.

-----

env vars:

```
tuber env set TUBER_OAUTH_REDIRECT_URL "https://admin.freshlyhq.com/tuber/auth" -a tuber -y
tuber env set TUBER_OAUTH_WEB_CLIENT_SECRET {web client from ui for each project} -a tuber -y
tuber env set TUBER_OAUTH_WEB_CLIENT_ID {web client from ui for each project} -a tuber -y
tuber env set TUBER_OAUTH_STATE_KEY asdfasdf -a tuber -y
```


pkg/adminserver/server.go
```diff
@@ -84,10 +89,67 @@ func (s server) prefixed(route string) string {
...
+		w, r, authed, err = s.authenticator.TrySetCookieAuthContext(w, r)
+		if authed {
+			next.ServeHTTP(w, r)
+			return
+		}
+
+		http.Redirect(w, r, s.authenticator.RefreshTokenConsentUrl(), http.StatusUnauthorized)
+		if err != nil {
+			fmt.Println(err)
+		}
```
jaysonesmith, line 111: Same question here as on your other PR about -always- returning unauthed regardless of whether or not the `TrySetCookieAuthContext` errors

pkg/adminserver/server.go
```diff
@@ -84,10 +89,67 @@ func (s server) prefixed(route string) string {
...
+}
+
+func (s server) receiveAuthRedirect(w http.ResponseWriter, r *http.Request) {
+	queryVals := r.URL.Query()
+	if queryVals.Get("error") != "" {
+		http.Redirect(w, r, fmt.Sprintf("/tuber/unauthorized/&error=%s", queryVals.Get("error")), http.StatusUnauthorized)
+		return
+	}
+	if queryVals.Get("code") == "" {
+		http.Redirect(w, r, fmt.Sprintf("/tuber/unauthorized/&error=%s", "no auth code returned from iap"), http.StatusUnauthorized)
```
jaysonesmith, line 122: Do we need the sprintf here? A single string seems suitable

jaysonesmith (approved):

pkg/adminserver/server.go
```diff
@@ -84,10 +92,69 @@ func (s server) prefixed(route string) string {
...
+		r, authed = s.authenticator.TrySetHeaderAuthContext(r)
+		if authed {
+			next.ServeHTTP(w, r)
+			return
+		}
+
+		var err error
+		w, r, authed, err = s.authenticator.TrySetCookieAuthContext(w, r, s.secureCookie)
+		if err != nil {
+			http.Redirect(w, r, fmt.Sprintf("/tuber/unauthorized/&error=%s", err.Error()), http.StatusInternalServerError)
```
Hbbb, line 107: If we fail to set the cookie here, why don't we just redirect them back to the `RefreshTokenConsentUrl` and have them try again?

pkg/oauth/oauth.go
```diff
@@ -0,0 +1,181 @@
...
+	r = r.WithContext(context.WithValue(r.Context(), accessTokenCtxKey, accessToken))
+	r = r.WithContext(context.WithValue(r.Context(), accessTokenExpirationCtxKey, accessTokenExpiration))
+
+	if refreshed {
+		encodedRefresh, err := sc.Encode(refreshTokenCookieKey(), refreshToken)
+		if err != nil {
+			return w, r, false, fmt.Errorf("encode refresh token cookie error: %v", err)
+		}
+
+		encodedAccess, err := sc.Encode(refreshTokenCookieKey(), accessToken)
```
Hbbb, line 135: Did you mean to use `accessTokenCookieKey` here?

pkg/adminserver/server.go
Hbbb: To expand on this further, `TrySetCookieAuthContext` has a ton of failure paths bubbling up from low level operations like decoding cookies etc.

If there are any issues decoding the cookies or with the result of refreshing a token, for example, maybe we should just remove the bad cookies and redirect the user to the consent screen again?

pkg/oauth/oauth.go
```diff
@@ -0,0 +1,181 @@
...
+
+func accessTokenCookieKey() string {
+	return "TUBERATOKEN"
+}
+
+func accessTokenExpirationCookieKey() string {
+	return "TUBERATOKENEXP"
+}
+
+type oauthCtxKey string
```
Hbbb, line 49: What's the purpose of this type? It doesn't seem like we're enforcing it anywhere.

Hbbb (changes requested):
I'd like to chat about some of the decisions around refreshing the auth stuff, but otherwise this looks good :)

Hbbb (approved):

PR #256, opened 2021-08-06
multiple sentries per resource

jkminneti:
`tuber/sentryUrl` will work fine for backwards compatibility
add as many as you want with that prefix

### DIFF
--- a/pkg/core/releaser.go
+++ b/pkg/core/releaser.go
@@ -423,10 +423,9 @@ func (r releaser) yamlToAppResource(yamls []string, data map[string]string) (app
 		for k, v := range parsed.Metadata.Annotations {
 			if strings.HasPrefix(k, "tuber/sentryUrl") {
 				url, ok := v.(string)
-				if !ok || url == "" {
-					continue
+				if ok && url != "" {
+					sentryUrls = append(sentryUrls, url)
 				}
-				sentryUrls = append(sentryUrls, url)
 			}
 		}
 
--- a/pkg/core/releaser.go
+++ b/pkg/core/releaser.go
@@ -561,7 +561,7 @@ func (r releaser) goWatch(resource appResource, timeout time.Duration, errors ch
 
 		for _, url := range resource.sentryUrls {
 			wg.Add(1)
-			go func(errors chan rolloutError, wg *sync.WaitGroup) {
+			go func(url string, errors chan rolloutError, wg *sync.WaitGroup) {
 				_, logger := resource.scopes(r)
 				healthy, message := monitor.Sentry(logger, url, r.sentryBearerToken, resource.watchDuration)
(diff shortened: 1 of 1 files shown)


pkg/core/releaser.go
```diff
@@ -554,11 +559,11 @@ func (r releaser) goWatch(resource appResource, timeout time.Duration, errors ch
 			wg.Done()
 		}(errors, wg)
 
-		if resource.sentryUrl != "" {
+		for _, url := range resource.sentryUrls {
 			wg.Add(1)
 			go func(errors chan rolloutError, wg *sync.WaitGroup) {
 				_, logger := resource.scopes(r)
-				healthy, message := monitor.Sentry(logger, resource.sentryUrl, r.sentryBearerToken, resource.watchDuration)
+				healthy, message := monitor.Sentry(logger, url, r.sentryBearerToken, resource.watchDuration)
```
jaysonesmith, line 566: [C2 - Potentially broken under certain circumstances](https://production.admin.freshlyservices.net/engineering-handbook/docs/processes/software-development/code-review/#c2---potentially-broken-under-certain-circumstances)

VS Code alerted me to the use of `url` here.

Screen Shot 2021-08-09 at 12 50 46 PM

I did a google and found [this](https://eli.thegreenplace.net/2019/go-internals-capturing-loop-variables-in-closures/) page which says that due to the way go's internals work, we'd only ever add the last url in `resource.sentryUrls` `len(resource.sentryUrls)` times.  AFAIK the fix would be to pass the url into the closure like below?

```go
			go func(errors chan rolloutError, wg *sync.WaitGroup, url string) {
				_, logger := resource.scopes(r)
				healthy, message := monitor.Sentry(logger, url, r.sentryBearerToken, resource.watchDuration)
				if !healthy {
					errors <- rolloutError{
						err:                fmt.Errorf("sentry monitoring found failure"),
						resource:           resource,
						monitorFail:        true,
						monitorFailMessage: message,
					}
				}
				wg.Done()
			}(errors, wg, url)
```

jaysonesmith (changes requested):
Just the one. :)

pkg/core/releaser.go
jaysonesmith: I've verified this locally with a test function similar to the current implementation and got:

```go
urls := []string{"url-one", "url-two", "url-three"}

// out:
// i: 2, url: "url-three"
// i: 2, url: "url-three"
// i: 2, url: "url-three"
```

jkminneti: YUUUUUP shoot nice catch

jkminneti: in fact how did neither goland nor staticcheck catch that wtf

jaysonesmith (approved):

PR #260, opened 2021-08-14
async env and local frontend

jkminneti:
```
tuber dev-gql-server (appName)
```
no env vars. 

it pulls the input app from the remote tuber into your local database. it also does _local user impersonation_.

oh and also app env is async, lauren did frontend

wild friday night


pkg/k8s/configResource.go
```diff
@@ -75,3 +75,36 @@ func GetConfigResource(name string, namespace string, kind string) (config *Conf
 
 	return
 }
+
+// GetConfigResourceWithToken is a step toward a larger refactor
```
jkminneti, line 79: killed this in the subsequent pr, #262

pkg/adminserver/web/pages/apps/[id].tsx
```diff
@@ -227,13 +228,15 @@ const ShowApp = () => {
...
-					useSet={useSetAppEnvMutation}
-					useUnset={useUnsetAppEnvMutation}
-				/>
-			</Collapsible>
+			<CollapsedAsyncTextInputGroup 
+				tableHeading={'Environment Variables'} 
+				appName={app.name}
+				queryDocument={GetAppEnvDocument}
+				queryVars={{ name: app.name }}
+				queryData={function(data) { return data.getAppEnv}}
```
Comment (jkminneti) [inline pkg/adminserver/web/pages/apps/[id].tsx:236]: is this an absolute abuse of the any type? probably

Comment (Hbbb) [reply pkg/adminserver/web/pages/apps/[id].tsx:236]: Does this even need to exist? Why can't you just access `data.getAppEnv` directly inside of `CollapsedAsyncTextInputGroup`?

Comment (Hbbb) [reply pkg/adminserver/web/pages/apps/[id].tsx:236]: Actually I take that back. The fact that the accessor is a function makes this more generic. Nice!

Hbbb (approved):

cmd/env.go
```diff
@@ -256,9 +256,7 @@ func getAllEnvGraphqlQuery(appName string) (map[string]string, error) {
 		}
 	`
 
-	var respData struct {
-		GetApp *model.TuberApp
-	}
+	var respData []*model.Tuple
```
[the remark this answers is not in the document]
jkminneti: turns out that broke it tho :( it's reverted and fixed in a subsequent pr

PR #261, opened 2021-08-17
Add Cloud Build Notifications to Tuber

Hbbb:
Depends on #268 

## What does this PR do?
Adds Cloudbuild notifications to Tuber!

Tuber will now notify build starts, successes and failures, with links to the cloud build run.

Slack___lauren-sneaky-channel___Freshly

## How do I test this?
Use the `startBuilds` test command. That will start up _only_ the cloud build listener, and also populate a local boltDB with the potatoes app.

I've been using [this build](https://console.cloud.google.com/cloud-build/builds;region=global/ac33d997-1f09-4f81-8556-a9fe6f942950?project=freshly-docker) (just keep rebuilding it) to test. Alerts go to #lauren-sneaky-channel.

## Code Thoughts
The most significant change from a code perspective was updating `pubsub.Listener` to take a `Processor` interface, rather than a concrete type. The changes required for this are pretty minimal. 

## Concerns
Currently this will send build notifications to the same channel as release notifications. There are two issues here: it's noisy and the messages can sometimes come in out of order. We may consider adding an additional slack channel field to the apps in bolt for build notifications.

## Deployment & Testing
Set these new env vars:
```bash
tuber env set -a tuber TUBER_PUBSUB_PROJECT freshly-docker
tuber env set -a tuber TUBER_PUBSUB_CLOUDBUILD_SUBSCRIPTION_NAME tuber-cloud-builds
```

Deploy this to staging for a day or two and see how things shake out.
```bash
tuber apps set branch tuber cloudbuild-notifications
tuber deploy tuber
```

## TODO
- [ ] Disable link unfurling in tuber's slack client [API Docs](https://api.slack.com/reference/messaging/link-unfurling#classic_unfurl) (#268)


pkg/builds/listener.go
```diff
@@ -0,0 +1,93 @@
...
+		return
+	}
+
+	if len(event.Images) < 1 {
+		event.Logger.Debug("build contains no images; skipping")
+		return
+	}
+
+	var apps []*model.TuberApp
+	for _, img := range event.Images {
```
Hbbb, line 68: I'm a bit concerned about this. I can't imagine how this could ever result in multiple hits for the same app, but if it did, it would send duplicate notifications.

pkg/events/processor.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/freshly/tuber/graph/model"
 	"github.com/freshly/tuber/pkg/core"
 	"github.com/freshly/tuber/pkg/gcr"
+	psub "github.com/freshly/tuber/pkg/pubsub"
```
Hbbb, line 15: I think it makes sense to alias our own `pubsub` package due to the tiny amount of usages in this package. Happy to discuss though.

cmd/postmsg.go
```diff
@@ -0,0 +1,64 @@
+package cmd
```
Hbbb, line 1: This file was just for debugging. I'll delete it before I merge.

cmd/start.go
```diff
@@ -100,6 +101,24 @@ func start(cmd *cobra.Command, args []string) error {
...
+		data,
+		buildEventProcessor,
+	)
+	if err != nil {
+		startupLogger.Error("failed to start cloud build listener", zap.Error(err))
+		report.Error(err, scope.WithContext("initialize cloud build listener"))
+		panic(err)
+	}
+
+	go buildListener.Start()
```
Hbbb, line 120: The `Start` command logs when errors happen. Since this isn't critical path, we don't really care if this fails to start. We definitely don't want to halt the program, right? Open to discussion on how to handle errors here.

cmd/startBuilds.go
```diff
@@ -0,0 +1,100 @@
+package cmd
```
Hbbb, line 1: This file was just for testing/debugging. I'll delete it before I merge.

jkminneti (commented):
such a huge feature, so little code ❤️

mostly small things

cmd/start.go
```diff
@@ -100,6 +101,24 @@ func start(cmd *cobra.Command, args []string) error {
...
+		viper.GetString("TUBER_PUBSUB_PROJECT"),
+		viper.GetString("TUBER_PUBSUB_CLOUDBUILD_SUBSCRIPTION_NAME"),
+		creds,
+		data,
+		buildEventProcessor,
+	)
+	if err != nil {
+		startupLogger.Error("failed to start cloud build listener", zap.Error(err))
+		report.Error(err, scope.WithContext("initialize cloud build listener"))
+		panic(err)
```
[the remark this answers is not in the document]
jkminneti: actually - you know what - it's not. not without wrapping start and passing errors to sentry. this is correct but _just for this command okaaay_

pkg/builds/listener.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	return matches, nil
+}
+
+func buildMessage(event *Event, app *model.TuberApp) string {
+	var msg string
+	switch event.Status {
+	case "WORKING":
+		msg = fmt.Sprintf(":building_construction: Build started for *%s* - <%s|Build Logs>", app.Name, event.LogURL)
```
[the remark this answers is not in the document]
jkminneti: also give this a try if you want
```
slack.MsgOptionDisableLinkUnfurl()
```

pkg/builds/listener.go
```diff
@@ -0,0 +1,120 @@
...
+	if event.Substitutions.BranchName == "" {
+		event.logger.Debug("build notification payload missing substitutions.BRANCH_NAME")
+		return
+	}
+
+	// This code is to handle Github App triggers. In those cases, event.Substitutions.RepoName is just the app name i.e. "distribution"
+	// In Tuber's DB we store all cloudSoureRepos as "github_freshly_{appName}"
+	repoName := event.Substitutions.RepoName
+	if !strings.Contains(event.Substitutions.RepoName, "github_freshly") {
+		repoName = "github_freshly_" + event.Substitutions.BranchName
```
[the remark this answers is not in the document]
Hbbb: I make the repoName look like a cloud source repo in case it's a Github App, in which case it would be `distribution`. 

So in this example, it'd see `event.Substitutions.RepoName == "ditribution"` and overwrite it to be `"github_freshly_distribution"`. That's how it's stored in the DB.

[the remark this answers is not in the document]
jkminneti: sure but we can't hardcode `github_freshly_`

pkg/builds/listener.go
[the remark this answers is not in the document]
Hbbb: How do you know what the standard library of slack emojis is?

pkg/builds/listener.go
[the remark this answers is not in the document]
jkminneti: i think the easiest path forward would be to match on repo name OR app name

pkg/builds/listener.go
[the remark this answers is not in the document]
jkminneti: https://www.webfx.com/tools/emoji-cheat-sheet/

but oops

cus i just looked manually and it's totally in there.

literally 
https://slack.com/help/articles/202931348-Use-emoji-and-reactions
they told me to go there 😭

pkg/events/processor.go
jaysonesmith: This brings up an interesting thought on how to handle conflicts. Do we say whatever conflicting package is used the _least_ in a file gets aliased? Do we say that external packages get aliased first or the inverse?

If frequency based, how do we stay consistent across the wider project where usage of the same packages may be swapped?

🤔 

My personal immediate feeling would be to prioritize aliasing external packages first. It could theoretically help signal to readers that because it's been aliased, it's not our package/code?

cmd/start.go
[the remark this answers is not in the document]
Hbbb: 👍

pkg/pubsub/pubsub.go
```diff
@@ -23,12 +22,16 @@ type Listener struct {
 	subscriptionName string
 	credentials      []byte
 	clusterData      *core.ClusterData
-	processor        *events.Processor
+	processor        Processor
+}
+
+type Processor interface {
```
jaysonesmith, line 28: I'm not super familiar with tuber's pubsub stuff in its various places, but based on the fact that we've got three different declarations of a type `Processor` in the project, that we should probably rethink the structure/naming/usage of things as a whole. Not specifically saying your code or for this PR, just noting things out loud and getting it in our heads. haha

(in order they appear in my IDE)
1. `Processor` struct in `./pkg/builds/listener.go` - Probably makes sense to have the exported object here actually be a `Listener` struct and have `Process` be a method attached to it like we currently have with `ProcessMessage` - Though I don't think this is the place this should necessarily live either.
2. `Processor` struct in `./pkg/events/processor.go` - Based on Go's naming schemes, this would likely turn into `Eventer`. I'm not entirely sure that the `events` package is the correct place for all the code in this file though. Outside its `ProcessMessage` method it exports `ReleaseApp` and `StartRelease`. Based on their similar naming it feels like they should be in a `releases` package with a `Releaser` struct/interface.
3. `Processor` interface in `./pkg/pubsub/pubsub.go` - IMO, our `pubsub` package should be the place where we have files for listening for and processing events, we could create interfaces for them here, export them, and then anything in tuber that needs to deal with pubsub stuff can utilize them.

Wider scope than this PR, just food for thought. 🪙

pkg/pubsub/pubsub.go
```diff
@@ -23,12 +22,16 @@ type Listener struct {
...
+	ProcessMessage(Message)
 }
 
 // NewListener is a constructor for Listener with field validation
 func NewListener(ctx context.Context, logger *zap.Logger, pubsubProject string, subscriptionName string,
-	credentials []byte, clusterData *core.ClusterData, processor *events.Processor) (*Listener, error) {
+	credentials []byte, clusterData *core.ClusterData, processor Processor) (*Listener, error) {
```
jaysonesmith, line 34: 👍 for accepting interfaces but returning structs!

pkg/pubsub/pubsub.go
```diff
@@ -50,19 +53,24 @@ func NewListener(ctx context.Context, logger *zap.Logger, pubsubProject string,
...
 	Tag    string `json:"tag"`
+
+	// Cloud Build Pubsub message attributes
+	Status        string `json:"status"`
+	LogURL        string `json:"logUrl"`
+	Substitutions struct {
+		BranchName string `json:"BRANCH_NAME"`
+		RepoName   string `json:"REPO_NAME"`
+	} `json:"substitutions"`
 }
```
jaysonesmith, line 69:
```suggestion
// Message is the JSON deserialization target for pubsub messages
type Message struct {
	// GCR Pubsub message attributes
	Digest string `json:"digest"`
	Tag    string `json:"tag"`

	// Cloud Build Pubsub message attributes
	Status        string         `json:"status"`
	LogURL        string         `json:"logUrl"`
	Substitutions *Substitutions `json:"substitutions"`
}

type Substitutions struct {
	BranchName string `json:"BRANCH_NAME"`
	RepoName   string `json:"REPO_NAME"`
}
```

Embedded anonymous structs make things difficult to work with and prevent re-usability elsewhere. If we wanted to create an object of type `Message` as it stands now, we'd have to declare the entire anonymous struct right down to the json annotations in order for the compiler to be cool with it.

```go
msg := Message{
	Digest: "foo-digest",
	Tag:    "foo-tag",
	Status: "foo-status",
	LogURL: "foo-logurl",
	Substitutions: struct {
		BranchName string `json:"BRANCH_NAME"`
		RepoName   string `json:"REPO_NAME"`
	}{
		BranchName: "foo-branch",
		RepoName:   "foo-repo",
	},
}
```

The suggested change would allow for this:

```go
msg := Message{
	Digest: "foo-digest",
	Tag:    "foo-tag",
	Status: "foo-status",
	LogURL: "foo-logurl",
	Substitutions: Substitutions{
		BranchName: "foo-branch",
		RepoName:   "foo-repo",
	},
}
```

While yes, it's a small improvement, memory allocation/usage can be improved by utilizing pointers to the `Substitutions` struct. Without a pointer, an empty instantiated message struct is 96 bytes, with the substitutions struct being 32. With a pointer the numbers drop to 72 and 8 for message and substitutions, respectively. IMO, pointers should always be used in this type of situation for consistency, speed, etc.

I'd also recommend adding `,omitempty` to the json annotation for Substitutions unless having an empty object is always needed if marshaling to json? If switching to the pointer usage, then an empty message turns into json with a null field. (which may be desired 🤷 )

```json
{"substitutions":null,"digest":"","tag":"","status":"","logUrl":""}
```

With `omitempty`

```json
{"digest":"","tag":"","status":"","logUrl":""}
```

Thanks for coming to my TED talk.

jaysonesmith: When looking into this, I also ran across this [blog post](https://medium.com/techverito/golang-struct-size-and-memory-optimisation-b46b124f008d) which I think is pretty informative inre the ordering of elements of a struct and allocation. LMK what you think of it, too.

Hbbb: Alright. I'm down with these changes, but it's worth pointing out that `pubsub.Message` is used solely to deserialize, not to serialize. It's unlikely we'll need to create instances of it directly, except maybe in test files 😅. Regardless, this is a good suggestion and I'll do it.

cmd/start.go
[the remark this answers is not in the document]
jaysonesmith: Why are we using panic at all? I disagree with the use of all of them in this file. Any error in this file would be returned to the `main` func in the tuber's root `main.go` file which handles it with `log.Fatal(err)`.

There's also [uber's take](https://github.com/uber-go/guide/blob/master/style.md#dont-panic) on panics. tl;dr - don't. haha

pkg/pubsub/pubsub.go
Hbbb: Definitely agree about the need for a `releases` package. The `events` package is one of the original packages and it clearly needs to be renamed. One of my goals with this PR was to minimize the impact radius of this change though, so I didn't do any of that.

Let me address your points in order:
1. I renamed the file to `events.go` and kept it as a `Processor`. The reason is because it's explicitly not a listener, it is a processor. It receives messages and does a thing. I quite like this separation between the `pubsub.Listener`, which takes a generic processor and passes it messages.
2. Couldn't agree more. We need a `releases` package which more accurately reflects the purpose of the package. I can see that package containing an `events.go` file that implements the `Processor` interface, which would contain much of what we see in `events/processor.go` today.
3. I'm actually pretty happy with the `pubsub` package. The fact that I had to change so little code to spin up another event processor is a testament to it's design, IMO. We're already exporting an interface that effectively says: "If you want to receive pubsub messages, pass me something with a `Process` method and I'll pass you the event". I am open to more discussion on it though.

In terms of actionable feedback on this PR, is there anything you'd like to see changed before we merge this?

jaysonesmith: 1. 👍 You know I'm all about separation of concerns so no push back from me here!
2. Yup! May not even need an events file specifically; that's something that the code can tell us when writing it though.
3. My take here is that having a set home for the baseline code for dealing with pubsub stuff. (like where I'd like the interface to live)

Unrelated to the PR and I'm not expecting you to know (yay if you do), I'm curious as why we're not utilizing the same nomenclature of the pubsub world: subscription, topic, publish, etc.

jaysonesmith: Totally missed your last question: Nope! ✅ incoming

jaysonesmith (approved):

pkg/pubsub/pubsub.go
jaysonesmith: Ah yea, that's fair. You **know** I want tests for as much as we can get, sooooo .... hahaha

pkg/pubsub/pubsub.go
Hbbb: > Unrelated to the PR and I'm not expecting you to know (yay if you do), I'm curious as why we're not utilizing the same nomenclature of the pubsub world: subscription, topic, publish, etc.

There's not a good answer here. I think at the time it made more sense for the pubsub thing to be a "Listener" vs a subscription? But none of this stuff is set in stone of course. Lets change it if it will make the code more comprehensible.

PR #262, opened 2021-08-23
simple auth checks

jkminneti:
https://freshly.atlassian.net/browse/POPS-1206

add simple can-i's to every resolver. full user impersonation can come later once the kubectl funcs are refactored to use better inputs


graph/resolver.go
```diff
@@ -35,3 +40,65 @@ func NewResolver(db *core.DB, logger *zap.Logger, processor *events.Processor, c
...
+	token, err := oauth.GetAccessToken(ctx)
+	if err != nil {
+		return fmt.Errorf("error retrieving authorization params")
+	}
+	authorized, err := k8s.CanI(appName, verb, subject, "--token="+token)
+	if err != nil {
+		return fmt.Errorf("error determining authorization status")
+	}
+	if !authorized {
+		return fmt.Errorf("unauthorized to perform this action")
```
Hbbb, line 101: Since we have the verb and subject, why not interpolate that into the error for a more descriptive error message for the user?

graph/schema.resolvers.go
```diff
@@ -219,6 +267,11 @@ func (r *mutationResolver) UnsetAppEnv(ctx context.Context, input model.SetTuple
 }
 
 func (r *mutationResolver) SetExcludedResource(ctx context.Context, input model.SetResourceInput) (*model.TuberApp, error) {
+	err := canDeleteDeployments(ctx, input.Name)
```
Hbbb, line 270: Interesting permission check here. I guess it makes sense, since the excluded resources won't be created?

Hbbb (approved):
Small suggestion on better error messages.

PR #267, opened 2021-09-22
Use Docker Scratch Image

Hbbb:
First off, no one asked for this and we absolutely don't need to do this.

## What does this do?
* Updates our `Dockerfile` to use stages (that new fancyness)
* Updates tuber to run on a [scratch](https://hub.docker.com/_/scratch) image. It's super barebones

## Notes
* I had to run `dos2unix` on the entire project because we had hidden Windows characters in our files and it was causing weird issues
* I needed to disable CGO via `CGO_ENABLED=0` 

The second point requires more explanation. [This](https://stackoverflow.com/questions/61515186/when-using-cgo-enabled-is-must-and-what-happens) Stackoverflow post goes into good detail about what CGO is and why it's sometimes necessary. I wasn't even able to run `tuber --help` without this setting. 

I found the solution from [this](https://old.reddit.com/r/docker/comments/bld85r/help_exec_user_process_caused_no_such_file_or/) Reddit thread.


Hbbb:
@jkminneti Brought up a great point: With this change we won't be able to `bash` or even `sh` into a running tuber image. That's enough of a reason not to go through with this. 

It was a fun experiment.

PR #269, opened 2021-09-23
Don't Query for Builds with no (build) trigger ID

Hbbb:
* Early exit from finding builds if a `TuberApp` doesn't have a `triggerID`

Question(s):
* Should the absence of a `triggerID` in `builds.FindByApp` return an error, or just an empty result set?

### DIFF
--- a/pkg/builds/builds.go
+++ b/pkg/builds/builds.go
@@ -22,6 +22,10 @@ func FindByApp(app *model.TuberApp, triggersProjectName string) ([]*model.Build,
 		return nil, err
 	}
 
+	if app.TriggerID == "" {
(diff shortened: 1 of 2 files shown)


pkg/builds/builds.go
```diff
@@ -22,6 +22,10 @@ func FindByApp(app *model.TuberApp, triggersProjectName string) ([]*model.Build,
 		return nil, err
 	}
 
+	if app.TriggerID == "" {
+		return make([]*model.Build, 0), nil
```
Hbbb, line 26: Should this return an error?

jkminneti (approved):

PR #270, opened 2021-09-25
review app reaper

jkminneti:


cmd/apps-info.go
```diff
@@ -44,6 +44,7 @@ func runAppsInfoCmd(cmd *cobra.Command, args []string) error {
 	table.SetRowLine(true)
 	table.SetAutoWrapText(false)
 	table.Append([]string{"Name", app.Name})
+	table.Append([]string{"Timestamps", "CreatedAt: " + app.CreatedAt + "\nUpdatedAt " + app.UpdatedAt})
```
jaysonesmith, line 47: I assume this is how we're writing to a table, yea? Is this writing to a single column called Timestamps with both the created & updated times within a single cell? If so, IMO we should have a specific column for created at and a column for updated at.

cmd/review-app-reaper.go
```diff
@@ -0,0 +1,85 @@
...
+				}
+
+				gql := `
+					mutation($input: AppInput!) {
+						destroyApp(input: $input) {
+							name
+						}
+					}
+				`
+				err = client.Mutation(context.Background(), gql, nil, input, &destroyResponse)
```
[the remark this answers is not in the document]
jaysonesmith:
```suggestion
		if err := client.Mutation(context.Background(), gql, nil, input, &destroyResponse); err != nil {
			return err
		}
```

cmd/review-app-reaper.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/spf13/viper"
+)
+
+var reviewAppReaperCmd = &cobra.Command{
+	SilenceUsage: true,
+	Hidden:       true,
+	Use:          "review-app-reaper",
+	Short:        "very internal, deletes old review apps",
+	Args:         cobra.NoArgs,
+	RunE: func(cmd *cobra.Command, args []string) error {
```
jaysonesmith, line 23: I really, really dislike having all of the behavior of this command in this spot. It means that it's basically 100% untestable.

jaysonesmith: It's also inconsistent with how we're writing other commands.

graph/schema.resolvers.go
```diff
@@ -824,6 +824,15 @@ func (r *queryResolver) GetApps(ctx context.Context) ([]*model.TuberApp, error)
 	return r.Resolver.db.SourceApps()
 }
 
+func (r *queryResolver) GetAllReviewApps(ctx context.Context) ([]*model.TuberApp, error) {
+	err := canViewAllApps(ctx)
+	if err != nil {
+		return nil, err
+	}
```
jaysonesmith, line 831:
```suggestion
	if err := canViewAllApps(ctx); err != nil {
		return nil, err
	}
```

cmd/apps-info.go
jaysonesmith: Nvm, just got to the schema and it looks separate. Guess it just reads weirdly.

jaysonesmith (approved):
Nothing appears broken. 😄

PR #271, opened 2021-09-27
save all apps

jkminneti:
the general migration tool

### DIFF
--- a/cmd/save-all-apps.go
+++ b/cmd/save-all-apps.go
@@ -3,8 +3,6 @@ package cmd
 import (
 	"context"
 
-	"github.com/freshly/tuber/graph/model"
-
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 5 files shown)


cmd/save-all-apps.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"context"
+
+	"github.com/spf13/cobra"
+)
+
+var saveAllAppsCmd = &cobra.Command{
+	SilenceUsage: true,
+	Hidden:       true,
+	Use:          "save-all-apps",
```
Hbbb, line 12: We could call this `migrate` and have the description explain what it does? Doesn't really matter that much though.

Hbbb (approved):

cmd/save-all-apps.go
jaysonesmith: I agree with this as "save-all-apps" isn't descriptive about what's actually happening.

graph/schema.resolvers.go
```diff
@@ -755,6 +755,28 @@ func (r *mutationResolver) UnsetRacExclusion(ctx context.Context, input model.Se
 	return app, nil
 }
 
+func (r *mutationResolver) SaveAllApps(ctx context.Context) (*bool, error) {
+	// just to keep it to super admins and tuber itself
+	err := canCreateApps(ctx)
+	if err != nil {
+		return nil, err
+	}
```
jaysonesmith, line 763:
```suggestion
	if err := canCreateApps(ctx); err != nil {
		return nil, err
	}
```

cmd/save-all-apps.go
```diff
@@ -0,0 +1,34 @@
...
+	Short:        "general migration tool - internal, hidden, but also, optimistically, always safe",
+	Args:         cobra.NoArgs,
+	PreRunE:      promptCurrentContext,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		graphql, err := gqlClient()
+		if err != nil {
+			return err
+		}
+		var respData interface{}
```
jaysonesmith, line 21: Two things: Can we declare this closer to its usage? Given how this function reads, does it mean that we're doing nothing with the response data that we're getting back?

graph/schema.resolvers.go
```diff
@@ -755,6 +755,28 @@ func (r *mutationResolver) UnsetRacExclusion(ctx context.Context, input model.Se
 	return app, nil
 }
 
+func (r *mutationResolver) SaveAllApps(ctx context.Context) (*bool, error) {
```
jaysonesmith, line 758: Can we drop the pointer to bool response value here? Every case in this function is returning a nil value and from what I can tell, it's not being used by anything calling it?

jaysonesmith (approved):

graph/schema.resolvers.go
jkminneti: nope the func defs are generated

PR #273, opened 2021-10-21
split create and run trigger

jkminneti:
long have we waited

### DIFF
--- a/graph/schema.resolvers.go
+++ b/graph/schema.resolvers.go
@@ -147,7 +147,7 @@ func (r *mutationResolver) CreateReviewApp(ctx context.Context, input model.Crea
 
 	sourceApp, err := r.Resolver.db.App(input.Name)
 	if err != nil {
-		return nil, fmt.Errorf("can't find source app. is %s managed by tuber", sourceApp)
+		return nil, fmt.Errorf("can't find source app. is %s managed by tuber", sourceApp.Name)
 	}
 
 	name, err := reviewapps.CreateReviewApp(ctx, r.Resolver.db, r.Resolver.logger, sourceApp, input.BranchName, input.Name, r.Resolver.credentials, r.Resolver.projectName)
--- a/graph/schema.resolvers.go
+++ b/graph/schema.resolvers.go
@@ -145,7 +145,12 @@ func (r *mutationResolver) CreateReviewApp(ctx context.Context, input model.Crea
 		return nil, err
 	}
 
-	name, err := reviewapps.CreateReviewApp(ctx, r.Resolver.db, r.Resolver.logger, input.BranchName, input.Name, r.Resolver.credentials, r.Resolver.projectName)
+	sourceApp, err := r.Resolver.db.App(input.Name)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


graph/schema.resolvers.go
```diff
@@ -145,7 +145,12 @@ func (r *mutationResolver) CreateReviewApp(ctx context.Context, input model.Crea
 		return nil, err
 	}
 
-	name, err := reviewapps.CreateReviewApp(ctx, r.Resolver.db, r.Resolver.logger, input.BranchName, input.Name, r.Resolver.credentials, r.Resolver.projectName)
+	sourceApp, err := r.Resolver.db.App(input.Name)
+	if err != nil {
+		return nil, fmt.Errorf("can't find source app. is %s managed by tuber", sourceApp.Name)
+	}
+
+	name, err := reviewapps.CreateReviewApp(ctx, r.Resolver.db, r.Resolver.logger, sourceApp, input.BranchName, input.Name, r.Resolver.credentials, r.Resolver.projectName)
```
mxygem, line 153: Would it make more sense and clean things up to just pass a pointer to the resolver itself? That would drop the argument count from eight to five.

Before
```go
CreateReviewApp(
	ctx context.Context, 
	db *core.DB, 
	l *zap.Logger, 
	sourceApp *model.TuberApp, 
	branch string, 
	appName string, 
	credentials []byte, 
	projectName string,
) (string, error)
```

After
```go
CreateReviewApp(
	ctx context.Context, 
	rslvr *graph.Resolver, 
	sourceApp *model.TuberApp, 
	branch string, 
	appName string,
) (string, error)
```

mxygem: Without knowing the variance of the arguments, it seems like it could be further simplified by dropping `appName` by utilizing `sourceApp.Name`? It seems like at this point we haven't begun to create the review app itself so we're still working with only the source app?

mxygem: Also, what's the difference between `sourceApp.Name` and `Resolver.projectName`?

pkg/reviewapps/reviewapps.go
```diff
@@ -75,10 +76,46 @@ func CreateReviewApp(ctx context.Context, db *core.DB, l *zap.Logger, branch str
...
+	allTriggers, listErr := call.Do()
+	if listErr != nil {
+		logger.Error("triggers list failed, skipping exists check")
+	} else {
+		for _, trigger := range allTriggers.Triggers {
+			if trigger.Name == reviewAppName {
+				triggerID = trigger.Id
+			}
+		}
+	}
```
mxygem, line 101: This block seems like a decent candidate to move into its own function encapsulating a particular set of responsibilities/actions.

It could theoretically also be split into three for further responsibility isolation: one that handles calling cloudbuild, one that looks for a match, and then one that calls both and is what is called here. Doing it this way could make testing easier due to the limited scope of each.

I really like what Quii wrote [here](https://quii.gitbook.io/learn-go-with-tests/questions-and-answers/os-exec) in response to a question they got.

PR #276, opened 2021-12-14
Update env file

caelra:
Update `env file` to allow the setting of multiple env variables through a yaml file, the old behavior used to override the current env variables config.

Now we can use the **replace flag** `tuber env file env.yaml --replace` on the env file command to override the current env variables configuration, if we use the default `tuber env file env.yaml` it will just add the env variables to our configuration.

### DIFF
--- a/cmd/env.go
+++ b/cmd/env.go
@@ -67,13 +67,12 @@ var fileCmd = &cobra.Command{
 			filePath = args[0]
 		}
 
-		switch replaceFlag {
-		case true:
+		if replaceFlag {
 			err := k8s.CreateEnvFromFile(appName, filePath)
 			if err != nil {
 				return err
 			}
-		default:
+		} else {
 			err := k8s.SetEnvFromFile("deployments", appName, filePath)
 			if err != nil {
 				return err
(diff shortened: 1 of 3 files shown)


cmd/env.go
```diff
@@ -65,10 +66,18 @@ var fileCmd = &cobra.Command{
 			appName = appNameFlag
 			filePath = args[0]
 		}
-		err := k8s.CreateEnvFromFile(appName, filePath)
 
-		if err != nil {
-			return err
+		switch replaceFlag {
```
Jefferson-Faseler, line 70: The replaceFlag var is a bool so I'm not sure how much of a switch statement is needed. Any reason you decided to go with that route?

pkg/k8s/secrets.go
```diff
@@ -135,3 +155,23 @@ func RemoveSecretEntry(mapName string, namespace string, key string) (err error)
 func CreateEnv(appName string) error {
 	return Create(appName, "secret", "generic", appName+"-env")
 }
+
```
Jefferson-Faseler, line 158: We should add a comment here for the documentation

pkg/k8s/secrets.go
```diff
@@ -118,6 +118,26 @@ func PatchSecret(mapName string, namespace string, key string, value string) (er
 	return config.Save(namespace)
 }
```
Jefferson-Faseler, line 120: This needs a documentation comment

cmd/env.go
```diff
@@ -275,6 +284,7 @@ func init() {
 	envUnsetCmd.Flags().StringVarP(&appNameFlag, "app", "a", "", "app name")
 	envCmd.AddCommand(envUnsetCmd)
 	fileCmd.Flags().StringVarP(&appNameFlag, "app", "a", "", "app name")
+	fileCmd.Flags().BoolVarP(&replaceFlag, "replace", "r", false, "replace all env variables")
```
Jefferson-Faseler, line 287: It's really easy to remember this feature in the context of the PR, but I worry there might need to be a little more detail here. Maybe something like `replace all env variables with contents of file` or something like that? It let's the person know what it is being replaced with a little more easily.

Jefferson-Faseler (changes requested):
Great stuff! This will make it a lot easier to set multiple vars at a time! 

Requesting changes for the missing comments, the other details are not blocking.

Jefferson-Faseler (approved):
Awesome stuff. Works great when testing this

PR #280, opened 2022-02-22
Fix rollback

LLParse:
Emulate the functionality of `rollback undo` by updating tuber's in-memory previous/current state and persisting to disk.

### DIFF
--- a/pkg/core/releaser.go
+++ b/pkg/core/releaser.go
@@ -706,7 +706,7 @@ func (r releaser) rollback(appliedResources []appResource, cachedResources []app
 func (r releaser) rollbackResource(applied appResource, cached appResource) error {
 	var err error
 	if applied.supportsRollback() {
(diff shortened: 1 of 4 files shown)


pkg/core/releaser.go
```diff
@@ -704,13 +704,7 @@ func (r releaser) rollback(appliedResources []appResource, cachedResources []app
...
 
 func (r releaser) rollbackResource(applied appResource, cached appResource) error {
-	var err error
-	if applied.supportsRollback() {
-		err = k8s.RolloutUndo(applied.kind, applied.name, r.app.Name)
-	} else {
-		err = k8s.Apply(cached.contents, r.app.Name)
-	}
-
+	err := k8s.Apply(cached.contents, r.app.Name)
```
LLParse, line 707: `kubectl rollout undo` uses revision history stored in the Deployment resource, but Tuber is tracking resource changes so let's synchronize on Tuber's state tracking instead of using BOTH tuber state tracking and kube state tracking.

